PR #3, opened 2021-07-27
Version 1.1 test

lolchocotaco:
What does this do:
* Adds a deployAll script. 
* Adds mock contracts that replaces the `valueOf` method on the treasury.


lolchocotaco:
Going to merge this in so We can move the scripts/documentation to the V2 branch.

PR #7, opened 2021-08-17
Unflattening the Repo

lolchocotaco:
### There are no core logic changes!

Theres a lot of stuff going on so taking it slowly.
Currently only separating out the "common pieces" in interfaces & libraries. 
 We **cannot** drop-in replace openzepplin because there are some additional methods that were added for a few  contracts. When I get a chance I will dig into how we can reconcile that, but for now this is just restructuring the existing code. 

The _right_ way to do this would be to have a common set of interfaces/libraries, and extend those when needed for specific 
 contracts. 

Observations:
- Ownable is a generic contract, however its being redefined and the `onlyOwner` method is being rewritten to `onlyPolicy` or `onlyManager` depending on the case.  Makes simplifying more difficult.
- OHM & sOHM are `ERC20Permit` (which hasn't been adopted yet), but wsOHM is normal 'ol ERC20? Why?  Do we even need ERC20Permit? https://eips.ethereum.org/EIPS/eip-2612 <-- "not for general use"
- We've augmented safemath to handle sqrrts for the bonding calculator. Prevents drop in replace of OZ everywhere. 
- ~We use IERC20 to reference ERC20 tokens, however we rely on `decimals()` to be a method present on the contract. It is a convenience method that is added onto ERC20 and not a standard spec. This causes some issues when attempting to standardize IERC20 usage across the codebase. I'd recommend we keep it to ERC20 if we expect there to be a `decimals()` method. I had to change wsOHM to have decimals because of that.~ Casting to `IERC20Metadata` where needed
- We have 3 types of access control that more/less do the same  thing Ownable, Governable, and Guardable. The last two were sandwiched on top of each other in some contracts so i split them out. 
-  wEthBond is Ownable while the regular Bond Depository is Governable/Guardable. ^ related to the above point, does the same shit.

The code compiles correctly, haven't tested anything else out yet. Would love feedback on general direction & how we can clean this up better. 

TODO:
- [ ]  Write tests 
- [ ] Address the TODOs in the code & reduce "one-off" library/interface additions.


contracts/StandardBondingCalculator.sol
```diff
@@ -301,6 +56,6 @@ contract OlympusBondingCalculator is IBondingCalculator {
         } else {
             reserve = reserve0;
         }
-        return reserve.mul( 2 * ( 10 ** OHM.decimals() ) ).div( getTotalValue( _pair ) );
+        return reserve.mul( 2 * ( 10 ** IERC20Metadata(address(OHM)).decimals() ) ).div( getTotalValue( _pair ) );
```
lolchocotaco, line 59: Are these parenthesis balanced properly? :O

lolchocotaco:
DOING IT!

PR #12, opened 2021-08-20
Version 2 governance

lolchocotaco:


contracts/governnace/GovernorOHMegaInterfaces.sol
```diff
@@ -0,0 +1,194 @@
...
+    function GRACE_PERIOD() external view returns (uint);
+    function acceptAdmin() external;
+    function queuedTransactions(bytes32 hash) external view returns (bool);
+    function queueTransaction(address target, uint value, string calldata signature, bytes calldata data, uint eta) external returns (bytes32);
+    function cancelTransaction(address target, uint value, string calldata signature, bytes calldata data, uint eta) external;
+    function executeTransaction(address target, uint value, string calldata signature, bytes calldata data, uint eta) external payable returns (bytes memory);
+}
+
+/// @notice change from original contract
+interface wsOHMInterface {
```
sali0, line 185: IMO this would be worth putting in its own file and include all of the special functions (wrapping, unwrapping, etc). wsOHM will be used in other contracts as well, so I think having a single file with its interface would be helpful for people trying to build on top of Olympus.

contracts/governnace/GovernorAlpha.sol
```diff
@@ -0,0 +1,386 @@
+pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
+
+contract GovernorAlpha {
+    /// @notice The name of this contract
+    string public constant name = "Olympus Governor Alpha";
```
sali0, line 6: Should this be "Olympus Governor OHMega" ? 😎

sali0 (commented):
Looks good, just small suggestions on naming and the wsOHM interface.

contracts/governnace/GovernorOHMegaInterfaces.sol
lolchocotaco: Agreed Good call!. we're currently unflattening the rest of the repo and we can pull out utility methods appropriately as well.

contracts/governnace/GovernorAlpha.sol
lolchocotaco: yeeep

PR #17, opened 2021-09-24
Tyche

ind-igo:


lolchocotaco:
is this.. the real branch?

ind-igo:
> 
> 
> is this.. the real branch?

Yessir. On current version of Version-2 branch

test/YieldDirector.test.js
```diff
@@ -0,0 +1,414 @@
...
+        await ohm.setVault(treasury.address)
+
+        // Set index to 10
+        await sOhm.setIndex("10000000000");
+        // Initialization for sOHM contract.
+        await sOhm.initialize(staking.address);
+
+        // Set treasury, distributor, warmup, and locker for staking contract
+        await staking.setContract("0", distributor.address);
+        await staking.setContract("1", warmup.address);
```
naps62, line 94: randomly stumbled upon this. `setContract`, according to the code, is expecting to receive the distributor (0), and gOHM (1), but a warmup contract is being sent here?

possibly inconsequencial for this spec, but I'd like to understand. maybe I'm missing something obvious

ind-igo: I copied this from the old deploy script from v1.1 and since this was for tests, it didn't cause any issues. But good catch, will change this when i circle back to Tyche.

PR #44, opened 2021-10-23
We have tests now, let's run them in CI!

tonypiano:
This works by copying .env_sample to .env before running the tests. Not sure if that is what we were planning on doing in CI or not.


lolchocotaco:
Addressing this in : https://github.com/OlympusDAO/olympus-contracts/pull/57

PR #50, opened 2021-10-26
dapptools integration sample

kushal256:
This is a sample of how to test with dapptools.  Wanted to get some feedback as there are a few concerns:

Pros:
1.  Written in solidity
2. It's fast
3. Fuzz testing (https://github.com/dapphub/dapptools/tree/master/src/dapp#property-based-testing) functionality helps expose edge cases.  See commented out function ( https://github.com/OlympusDAO/olympus-contracts/pull/50/files#diff-c9d3ab8a71d26efe521f0e975b885c7a91e977efc4ab0abde9481f8895c10429R49-R65 ), which was used to generate the test cases where values were 'statically' created.

Cons:
1. There is no way to validate event emissions (see https://github.com/dapphub/dapptools/issues/696 )
2. No way to set `msg.sender` which makes auth testing logic difficult
3. Mocking has been significantly improved by using with https://github.com/gnosis/mock-contract, but there is still one issue with getting SafeERC20 usage fully working


lolchocotaco (approved):
i like it. LFG.

0xOcnus:
Hey!

I don't want to necro this but it is (I believe) possible to set `msg.sender` by setting the `DAPP_TEST_CALLER` environment variable. Obviously this means that the tests need to be split up a bit more so I would agree that it is somewhat cumbersome.

kushal256:
Hey @0xOcnus I was trying to set that but note this comment from: https://github.com/OlympusDAO/olympus-contracts/pull/50/files#diff-fa6b0bfd0dd55bb27fedb067ccdd2df8908305ca02883a28246f2fefcd710358R34

Was having problems setting this value via that env var.  Despite difficulties with setting this value, still think it's possible to validate any modifiers that are trying to enforce permissions.  Please let me know if there is something specific you are looking for?

PR #56, opened 2021-10-29
Additional changes for v2 

ind-igo:


lolchocotaco (approved):
I think this is legit.

PR #58, opened 2021-10-31
Test the migration

tonypiano:
Covers migrateContracts, migrateToken and migrateLP.

Defund is still skipped.

Adds a call to `safeApprove` before calling deposit, as it was throwing
errors.

Makes TODOs for the rest of the assets in the treasury.


lolchocotaco:
Should we merge this in and unskip defund later?

tonypiano:
This is ready to merge. I've unskipped the test and disabled the timeout. Let me know if you want me to skip it again before I merge.

PR #59, opened 2021-11-01
Version 2 test bond

kushal256:
- Moves dapp lib and src code into test/dapp_test 
- Add basic bonding test case


kushal256:
Ready for review

test/dapp_test/src/util/MockContract.sol
```diff
@@ -0,0 +1,380 @@
+// SPDX-License-Identifier: UNLICENSED
+
+//Copied from https://github.com/gnosis/mock-contract/blob/b3db814ac41a3b632d8441cd1fbfbfce54e056f7/contracts/MockContract.sol but removed call "encodeWithSignature("updateInvocationCount" towards end
```
lolchocotaco, line 3: is there anyway we can include as dep? or is this better?

lolchocotaco (approved):
me thinks it looks good.

test/dapp_test/src/util/MockContract.sol
kushal256: The reason why this was done is I had to do a modification, there was a bug in the usage with https://github.com/OlympusDAO/olympus-contracts/blob/Version-2/test/dapp_test/src/BondDepository.t.sol#L227

Perhaps I should add that comment here, too.

Looking at the github history of MockContract.sol, they had a switch to enable/disable this behavior, but then removed it so it always occurs.  I need to explore why they did that, but the project hasn't been touched in ages.  So for now thought this was the simplest path.  The downside is no way to do Verification (i.e. verify if a mock'd method was called).

PR #63, opened 2021-11-02
Fix daptools CI

kushal256:
Change versions from specific hashes to release versions - now things work


lolchocotaco (approved):
How do we manage deps like this in the future?

PR #65, opened 2021-11-03
try doubling default timeout

tonypiano:
Moved the timeout up to the describe level, now mocha will wait long enough.


lolchocotaco (approved):
kek.

PR #86, opened 2021-11-09
moved hardhat deploy over

ind-igo:


lolchocotaco:
Deleting in favor of: v2-deploy

PR #88, opened 2021-11-10
Fixed up some of the busted tests

tonypiano:
The migrator tests are still broken though. `setVault` is no longer a function on `OHM`, but once I remove that call, the `migrateContracts` function fails because only the vault can mint.


test/staking/StakingDistributor.test.ts
```diff
@@ -360,154 +347,4 @@ describe("Distributor", () => {
             });
         });
       });
-
-      describe("setAdjustment", () => {
```
tonypiano, line 364: All these tests were somehow duplicated, but outside the describe block with the `beforeEach`. Probably a merge conflict gone wrong

PR #100, opened 2021-11-17
minor improvements

olyzeus:
- add index view
- add old sOHM to pull index + enable balanceFrom/balanceTo before migration
- add migrated bool to facilitate above
- update interface to reflect
- fix transferFrom


lolchocotaco (approved):
Seems legit. Could be better if we added a test for gOHM index getting. but thats fine for now.

PR #102, opened 2021-11-20
gOHM cleanup

ind-igo:
Inherit erc20

### DIFF
--- a/contracts/governance/gOHM.sol
+++ b/contracts/governance/gOHM.sol
@@ -1,19 +1,16 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.7.5;
 
-import "../libraries/SafeERC20.sol";
 import "../libraries/SafeMath.sol";
 import "../libraries/Address.sol";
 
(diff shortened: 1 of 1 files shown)


JExtor:
Upon any transfer, mint, or burn of tokens _beforeTokenTransfer() is called, inside this function _moveDelegates() was also called to move any delegates when any transfer action was done. This has been removed since it is inherited from ERC20, which does not call this _moveDelegates() function nor has access to it. Overriding it in the gOHM contract itself to call _moveDelegates() should fix this.

JExtor (commented):
Upon any transfer, mint, or burn of tokens _beforeTokenTransfer() is called, inside this function _moveDelegates() was also called to move any delegates when any transfer action was done. This has been removed since it is inherited from ERC20, which does not call this _moveDelegates() function nor has access to it. Overriding it in the gOHM contract itself to call _moveDelegates() should fix this.

JExtor:
Looks good

JExtor (commented):
Looks good

PR #112, opened 2021-11-26
update bond initialization

olyzeus:
decrease inputs for adding bonds
- takes oracle and computes bcv/initial debt
- removes inputted bcv or minimum opening price
- adds dao fee to teller
- updates teller interface

note: setDepository() function on teller exists for testnet and should be removed before production

### DIFF
--- a/contracts/BondDepository.sol
+++ b/contracts/BondDepository.sol
@@ -1,433 +1,487 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-pragma solidity ^0.7.5;
+pragma solidity 0.7.5;
 pragma abicoder v2;
 
 import "./libraries/SafeMath.sol";
-import "./libraries/FixedPoint.sol";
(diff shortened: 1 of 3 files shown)


ind-igo:
Comments from myself, Jeff and JFry:
- Does not compile ser
- IOracle interface does not match whats used in BondDepo (getPrice vs assetPrice)
- Oracle is always required per bond now?
- Why 8 decimals? If this is for chainlink, this needs to be changed in case other oracles are used
- SafeMode should be per bond, not global to the depo, right?
- Should SafeMode ever be off?
- Need to add checks in enableBond and deprecateBond if safeMode is enabled or not
- Line 281: 2 transfers can be jsut one, transfer directly to treasury
- Should rename BondDepository's Bond struct to be something different from the one in BondTeller
- BondTeller indexesFor array should not be a state variable. Instead, make a public view function that returns this same array on each call.

PR #113, opened 2021-11-29
Update StakingDistributor.sol

olyzeus:


lolchocotaco (approved):
Seems  like a safe fix.

PR #121, opened 2021-12-04
Added some tests. Looking good. Also made change to indexesFor() in B…

JExtor:
…ond Teller. Was giving an invlaid opcode


Faultier69:
hello i was trying out your deployAll script and it errors at line 109 because MockDAO is undefinded, i was wondering what this variable is used for, also do you have tips on how to set the numbers if i wanna deploy on avax?

PR #127, opened 2021-12-10
Treasury timelock

lolchocotaco:
**What dis do**:
- Allows us to enable assets without a timelock BEFORE initialization, so timelock is disabled by default on launch. 
- Similar effect to "onchaingovernance"

### DIFF
--- a/contracts/Treasury.sol
+++ b/contracts/Treasury.sol
@@ -427,7 +427,7 @@ contract OlympusTreasury is OlympusAccessControlled, ITreasury {
      * @notice disables timelocked functions
      */
     function disableTimelock() external onlyGovernor {
-        require(timelockEnabled == false, "timelock already disabled");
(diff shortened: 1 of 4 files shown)


contracts/Treasury.sol
```diff
@@ -364,21 +360,14 @@ contract OlympusTreasury is OlympusAccessControlled, ITreasury {
...
         }
         permissionQueue.push(
-            Queue({
-                managing: _status, 
-                toPermit: _address, 
-                calculator: _calculator, 
-                timelockEnd: timelock, 
-                nullify: false, 
-                executed: false
-            })
```
lolchocotaco, line 381: our priettierrc did this.

olyzeus (approved):
looks good, much more descriptive. corrected a false that should have been true, and standardized how the function documentation is formatted. good to merge.

PR #132, opened 2021-12-14
add liquidity migrator

olyzeus:
contract to migrate LPs going forward

added a view function for time until next epoch on staking

removed redundant bounty mint on distributor


lolchocotaco (approved):
While we do have `MigrateLP` in the token migrator. 

this accounts for the difference between v1 ohm and v2 ohm when exchanging. 

Double checked the contract addesses they look good.

PR #134, opened 2021-12-16
Create SupplyRebalancer.sol

olyzeus:
Fixes a short-term supply imbalance resulting from migration


lolchocotaco:
Can you provide a bit more context around when this will be used/ how it should be used? 
And fix the compiling pls.

PR #135, opened 2021-12-16
FXS allocator

tonypiano:
Adds an FXS allocator that will lock FXS to veFXS for the maximum duration of 4 years. When yield is harvested, it is again locked and will extend the lock to 4 years if possible.

This contract is upgradable because veFXS is non-transferrable and we will want to use it for voting at some point soon. We can then upgrade this contract to include the voting logic.

This PR includes an integration test that forks mainnet and puts the allocator through its paces.


contracts/allocators/FraxSharesAllocator.sol
```diff
@@ -0,0 +1,172 @@
...
+contract  FraxSharesAllocator is Initializable, OwnableUpgradeable {
+    using SafeERC20 for IERC20;
+    using SafeMathUpgradeable for uint256;
+
+    /* ======== STATE VARIABLES ======== */
+    /* !!!! UPGRADABLE CONTRACT !!!! */
+    /* NEW STATE VARIABLES MUST BE APPENDED TO END */
+
+    uint256 constant private MAX_TIME = 4 * 365 * 86400 + 1;  // 4 years and 1 second
+    ITreasury public treasury;
```
lolchocotaco, line 86: So since FXS is in the v1 treasury, can we add a setter to this? should allow for EZ switching to v2 treasury when we done.

tonypiano: Donezo

ind-igo (approved):

PR #138, opened 2021-12-24
refactor

olyzeus:
merge depository and teller to save gas
create directory to ease deployment

let this be the last, inshallah


olyzeus:
dapp tests failing due to merge depo+teller, will circle back tomorrow if not taken care of

PR #145, opened 2021-12-30
V2 allocators convex

JExtor:


JExtor (commented):
Update looked good and clean up looks good. Thinking ready to go.

PR #147, opened 2022-01-02
Governance updates

JExtor:


contracts/governance/GovernorOHMegaInterfaces.sol
```diff
@@ -71,12 +71,12 @@ contract GovernorOHMegaDelegateStorageV1 is GovernorOHMegaDelegatorStorage {
     /// @notice The number of votes required in order for a voter to become a proposer
     uint public proposalThreshold;
 
-    /// @notice Initial proposal id set at become
-    uint public initialProposalId;
-
     /// @notice The total number of proposals
     uint public proposalCount;
 
+    /// @notice If has been initaited
```
lolchocotaco, line 77: So theres something with these `/// @notice`  blocks  within the governance folder. If you run solhint, it complains that the notices are never printed. Can you run this file through that and see what the case is?

PARABRAHMAN0:
Contact me if you want it reopened.

PR #149, opened 2022-01-05
REDACTED and LOBI staker

JExtor:


contracts/allocators/StakingManager.sol
```diff
@@ -0,0 +1,94 @@
...
+
+import "../types/OlympusAccessControlled.sol";
+
+interface IStaking {
+    function stake( uint _amount, address _recipient ) external returns ( bool );
+    function unstake( uint _amount, bool _trigger ) external;
+    function claim ( address _recipient ) external;
+}
+
+contract OlympusStakingManager is OlympusAccessControlled {
```
ind-igo, line 17: Should prob rename to something more specific. The name sounds general but this contract is specifically for Lobis and Redacted. Maybe `MetaGovernanceAllocator` ?

contracts/allocators/StakingManager.sol
```diff
@@ -0,0 +1,94 @@
...
+        require(address(authority.vault()) != address(treasury), "No change");
+        treasury = ITreasury(authority.vault());
+    }
+
+
+    /**
+     * @notice stakes either BTRFLY or LOBI from treasury
+     */
+    function stake(bool _redacted, uint _amount) external onlyGuardian {
+    (address staking, address token,) = _redactedOrLobi(_redacted);
```
ind-igo, line 45: indentation is off

ind-igo (changes requested):

contracts/allocators/StakingManager.sol
```diff
@@ -0,0 +1,94 @@
...
+        IStaking(staking).unstake(_amount, false);
+
+        // send token back to treasury
+        IERC20(token).safeTransfer(address(treasury), _amount);
+    }
+
+    /**
+     * @notice returns addresses depending on wanting to interact with redacted or lobi
+     */
+    function _redactedOrLobi(bool _redacted) internal view returns (address staking, address token, address stakedToken) {
```
lolchocotaco, line 82: I think this is okay for now. If we end up in a situation where we need another allocator for the same purpose, we should consolidate to a more generic implementation.

lolchocotaco (commented):
I think this is okay for now. If we end up in a situation where we need another allocator for the same purpose, we should consolidate to a more generic implementation.

contracts/allocators/StakingManager.sol
JExtor: done

contracts/allocators/StakingManager.sol
JExtor: done

contracts/allocators/StakingManager.sol
JExtor: Yea, was thinking the same thing, my thinking was that the big two we were wanting for BTRFLY and LOBI, wanted to make it where there was no address being passed in and only known addresses could be staked to with this implementation. Can always go back and deploy new contract if we come to situation where we are staking to other Olympus style protocols besides these two.

test/tokens/sOhm.test.ts
```diff
@@ -158,8 +159,11 @@ describe("sOhm", () => {
       });
     });
 
-    describe("circulatingSupply", () => {
+    // FIXME
```
lolchocotaco, line 162: What happened here?  Who gon fix. 
can also do `describe.skip`.

PR #150, opened 2022-01-07
add redirect deposit feature to tyche

Paul4912:
Motivation
Add function to allow users to redirect their deposit to different recipient

Changes
Added function to redirect deposit
Moved donationInfo[msg.sender].pop(); outside the if statement because it should be called regardless if deleted element happen to be at the end.
Changed arrays from storage to memory in the view functions

Have not added tests for my function, failing right now because rebase test is failing. somehow index goes from 10 to 20 in 1 rebase. any idea why @ind-igo


lolchocotaco:
The rebase test is broken. Could you dig into why?

PR #157, opened 2022-01-09
Bonding allocator

JExtor:


contracts/allocators/BondingAllocator.sol
```diff
@@ -0,0 +1,97 @@
...
+    function bond(
+        address _depository, 
+        address _token, 
+        uint256 _amount, 
+        uint256 _maxPrice, 
+        bool _managing
+    ) external onlyGuardian {
+        if(_managing) {
+            // retrieve amount of token from treasury
+            treasury.manage(_token, _amount);
```
lolchocotaco, line 64: Don't understand this `if` block. In what scenario would we want managing == false?

lolchocotaco (commented):
Overall looks okay. 
Could you add a little more context around which depositories we'd be bonding to? 

Is this essentially olympus treasury, to OP bonds?

contracts/allocators/BondingAllocator.sol
JExtor: Was thinking if there were tokens sent to the contract directly to be bonded. After second thought probably not needed and doesn't matter. Removed.

lolchocotaco:
Hey, closed the branch this PR was based off of, can you make it of of Main?

JExtor:
@lolchocotaco Off of main now

contracts/allocators/BondingAllocator.sol
```diff
@@ -0,0 +1,93 @@
...
+pragma solidity ^0.8.10;
+
+import "../libraries/SafeERC20.sol";
+
+import "../interfaces/IERC20.sol";
+import "../interfaces/ITreasury.sol";
+
+import "../types/OlympusAccessControlled.sol";
+
+interface IBond {
```
PARABRAHMAN0, line 11: Test

PR #159, opened 2022-01-10
Add FuseAllocator [WIP]

gregfromstl:


contracts/allocators/FuseAllocator.sol
```diff
@@ -0,0 +1,539 @@
...
+pragma solidity ^0.7.5;
+
+import "../libraries/SafeMath.sol";
+import "../libraries/SafeERC20.sol";
+import "../types/OlympusAccessControlled.sol";
+
+import "../interfaces/IERC20.sol";
+import "../interfaces/ITreasury.sol";
+
+interface ComptrollerInterface {
```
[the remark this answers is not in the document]
lolchocotaco: We've been saving the interfaces separately. +1 to moving it out of the core contract file.

contracts/allocators/FuseAllocator.sol
```diff
@@ -0,0 +1,539 @@
...
+        treasury.deposit(amount, tokenAddress, treasury.tokenValue(tokenAddress, amount));
+
+        return 0;
+    }
+
+    /**
+     * @notice Set the treasury address to be used.
+     * @param _treasury address The address of the Olympus treasury.
+     */
+    function setTreasury(address _treasury) public onlyGuardian returns (uint256) {
```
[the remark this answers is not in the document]
lolchocotaco: I don't think we need to do this. We should only be targeting the v2, treasury here so no need to update the treasury.

contracts/allocators/FuseAllocator.sol
```diff
@@ -0,0 +1,539 @@
...
+     * then enter the designated pool with those fTokens.
+     * @dev The amount of fTokens received is the number of tokens divided by the exchange rate
+     * (https://docs.rari.capital/fuse/#exchange-rate)
+     * @param comptrollerAddress address The address of the desired pool's comptroller.
+     * @param tokenAddress address The address of the token to deposit.
+     * @param fTokenAddress address The address of the fToken corresponding to the given token address.
+     * @param amount uin256 The amount of tokens to deposit.
+     */
+    function deposit(
+        address comptrollerAddress,
```
lolchocotaco, line 470: Weill these inputs change? 
Whats the strategy for the fuse allocator? any/all treasury assets or fixed ones?

contracts/allocators/FuseAllocator.sol
```diff
@@ -0,0 +1,539 @@
...
+
+        token.approve(fTokenAddress, amount); // approve the fToken address to transfer 'amount' in tokenAddress
+
+        CErc20Interface fToken = CErc20Interface(fTokenAddress);
+        uint256 mintAllowed = comptroller.mintAllowed(fTokenAddress, address(this), amount);
+        require(
+            mintAllowed == 0,
+            string(abi.encodePacked("Mint not allowed for specified token and amount, error code ", mintAllowed))
+        );
+        require(fToken.mint(amount) == 0, "Failed to mint fTokens");
```
lolchocotaco, line 487: Weird that it uses error codes?

contracts/allocators/FuseAllocator.sol
```diff
@@ -0,0 +1,539 @@
...
+
+        address[] memory fTokens;
+        fTokens[0] = fTokenAddress;
+        require(comptroller.enterMarkets(fTokens)[0] == 0, "Failed to enter pool");
+
+        return 0;
```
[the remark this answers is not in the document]
lolchocotaco: Seems like its following the error code pattern from fuse, but i don't like it. :(
Can we standardize on what we return?

contracts/allocators/FuseAllocator.sol
lolchocotaco: If its many assets, do you think its worth tracking which assets we're depositing?

lolchocotaco: Yeah taking a look at it, might be worth to have a mapping from token --> ftoken + comptroller (maybe)

contracts/allocators/FuseAllocator.sol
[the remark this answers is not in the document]
gregfromstl: Was hoping to not have 10 interface files for this single allocator. I guess that's okay with you all?

contracts/allocators/FuseAllocator.sol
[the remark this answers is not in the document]
gregfromstl: Yes was following Fuse's patterns. Can remove

contracts/allocators/FuseAllocator.sol
gregfromstl: Yeah I think that mapping makes sense, the issue is I don't know the fToken addresses yet as Rari's team keeps ghosting me when trying to troubleshoot their contracts. As for the inputs changing, I was told to make it adaptable just in case.

contracts/allocators/FuseAllocator.sol
gregfromstl: For anything that returns an error code yeah

contracts/allocators/FuseAllocator.sol
```diff
@@ -0,0 +1,539 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity ^0.7.5;
```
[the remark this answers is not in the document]
gregfromstl: It depends on OlympusAuthority which is locked to 0.7.5

contracts/allocators/FuseAllocator.sol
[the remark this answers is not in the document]
gregfromstl: I removed the unnecessary interface methods. Kept them in the file until they are needed by other contracts so as not to clutter the project, but a lot less interface fluff now

contracts/allocators/FuseAllocator.sol
```diff
@@ -0,0 +1,539 @@
...
+        require(
+            redeemAllowed == 0,
+            string(abi.encodePacked("Redeem not allowed for specified token and amount, error code ", redeemAllowed))
+        );
+        require(fToken.redeem(amount) == 0, "Failed to redeem fTokens");
+
+        IERC20 token = IERC20(tokenAddress);
+        token.approve(address(treasury), amount); // approve treasury to receive withdrawn tokens
+        // send all withdrawn tokens to treasury
+        treasury.deposit(amount, tokenAddress, treasury.tokenValue(tokenAddress, amount));
```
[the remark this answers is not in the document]
gregfromstl: Where can I determine if it's a reserve token?

contracts/allocators/FuseAllocator.sol
gregfromstl: I removed the return values on my methods. Do you want me to remove these requires as well? Personally, if they're using error codes I think we should keep these checks in but would love a second opinion.

contracts/allocators/FuseAllocator.sol
gregfromstl: I'm not exactly sure the range of assets that we'd want to use. My impression is for the foreseeable future it's just DAI. I could lock it in to be DAI only (we can withdraw and redeploy if we decide to expand) and maybe even default the comptroller to the Olympus Pool address?

contracts/allocators/FuseAllocator.sol
lolchocotaco: I think its fine to use them in the requires, you're checking that it was succesful, 
Is there any other communication of the error?

gregfromstl: You mean the specific error that the fuse contract is returning?

PR #160, opened 2022-01-10
comment out flaky assertion

tonypiano:


kushal256:
Wonder if this would improve if upgrading this contract to solidity > 0.8?

When in my LUSD Allocator PR, I bumped the solidity version of only my allocator (not Frax, not the treasury), I ran into the same problem and ended up solving by doing this: https://github.com/OlympusDAO/olympus-contracts/blob/v2-allocators-lusd4/test/allocators/FraxSharesAllocator.test.ts#L340

PR #164, opened 2022-01-12
Added CvxCRVAllocator.sol and Update IAllocator.sol

AoverK:
Added the Implementation contract for the cvxCRV Allocator.

Relies on a Proxy contract for reads and writes.

Also updated the IAllocator.sol Interface to contract to support both the current Frax Allocator Implementation contract and this cvxCRV Allocator Implementation contract.

The reason for updating IAllocator.sol is to start adding to the standardization of the Allocator Interface to support both current and future reward and token pools.


lolchocotaco:
Hey, closed the branch this PR was based off of, can you make it of of Main?

PARABRAHMAN0:
Old.

PR #180, opened 2022-01-18
Add profit when minting, so that the treasury has reserves

0xJem:


0xJem:
@lolchocotaco Minting fix, as we discussed earlier

lolchocotaco (approved):
Looks good, 1K profit should be good.

PR #183, opened 2022-01-18
LUSD Allocator:   Fix deadline

kushal256:


lolchocotaco (approved):
LGTM, assuming we don't have tests for this?

kushal256:
> LGTM, assuming we don't have tests for this?

Wish we did, it's difficult to test against a test network as the liquidity pools aren't the same between ropsten and mainnet.  

The other kind of testing would be mocking/verifying, but that's just going to validate a minimal piece of logic, which doesn't really offer much.

contracts/allocators/LUSDAllocator.sol
```diff
@@ -173,7 +173,7 @@ contract LUSDAllocator is OlympusAccessControlled {
                 ISwapRouter.ExactInputParams memory params = ISwapRouter.ExactInputParams({
                     path: abi.encodePacked(address(weth), poolFee, hopTokenAddress, poolFee, lusdTokenAddress),
                     recipient: address(this), //Send LUSD here
-                    deadline: block.timestamp + 25, //25 blocks, at 12 seconds per block is 5 minutes
+                    deadline: block.timestamp + 5 minutes,
```
tonypiano, line 176: You can just use `block.timestamp` the caller and the swap is in the same block.

I think the point of this is so you can't initiate a swap from the UI that sits in the mempool for days because your gas fee was too low and then you get rekt from a sandwich attack

tonypiano (changes requested):
small change

contracts/allocators/LUSDAllocator.sol
kushal256: Ahh i see what you mean now.  OK will fix it, don't think this will require a re-deployment of the contract then.

lolchocotaco:
What does deployment look like? can you update the PR with a deployment.json?

kushal256:
> What does deployment look like? can you update the PR with a deployment.json?

Do you mean a file like this?  https://github.com/OlympusDAO/olympus-contracts/blob/main/deployments/mainnet/gOHM.json

ind-igo:
This is already deployed, correct? I see the error, but this is incredibly small issue right? I don't think this is worth redeploying. Would rather wait for the new allocator system.

kushal256:
> This is already deployed, correct? I see the error, but this is incredibly small issue right? I don't think this is worth redeploying. Would rather wait for the new allocator system.

Agree - will close this then?

lolchocotaco:
Yeah favor of closeing then. Will revisit.

PR #185, opened 2022-01-19
style: format all files

mirru2532:
This PR formats all files ignoring 2 errors with ``GovernorOHMegaDelegate.sol`` and ``GovernorAlpha.sol``.

The important thing is here that when working on a fork users can now just ``yarn format:fix`` and subsequently ``git add .`` without having to worry if we for some reason do not want these formats (?)

State has been achieved by running above format command, once

Open for discussion


lolchocotaco:
YES! Thank you so much for this. can you add a `precommit` hook that performs the fix/lint? 
Also update the CI job?

mirru2532:
> YES! Thank you so much for this. can you add a `precommit` hook that performs the fix/lint? Also update the CI job?

Honestly I would just call ``yarn format`` from workflow instead of a precommit hook and make it clear for people to format their files, this should be common sense. Build simply fails if format wasn't called and we don't have to deal with hooks in that case.

lolchocotaco (approved):
Looks good to me and a great win for code standardization! 🎈

PR #186, opened 2022-01-19
feat: YieldStreamer and YieldSplitter Contract

Paul4912:
Motivation:
It's been requested in the discord for a contract to allow users to deposit their ohm and allow the yield to be converted to DAI in whatever time interval they want. The goal of the contract is to automate that and socialise the gas costs. 

Another goal is to encapsulate the common logic between yieldstreamer and yielddirector to create a reusable component that handles the logic for splitting deposits into the principal and yield component.

Changes:
Created mvp of the contract for review to get some feedback. If no glaring problems will proceed with tests and possibly engage frontend since we have idea of what interface they will call looks like.

For the common logic of splitting deposits into principal and yield I have created an abstract contract that serves as a base layer for all contracts that require this knowledge. The abstract contract is built to be as general as possible so it is easy to build on top of and make customisations. The contract aims to only concern itself with the accounting of splitting gOHM from its principal to yield component.

User Flow:
Users deposit their gOHM into the contract. They set parameters like userMinimumDaiThreshold(Lowest amount of dai they willing to accumulate before its send to them) and paymentInterval(minimum time elapse before their ohm yields are swapped into dai).

The DAO runs a gelato job so every once in a while(week maybe) it upkeeps the contract. During an upkeep users whos payment Interval has passed will have their yield in OHM swapped to DAI. If the amount of DAI they have to claim passes a user set threshold they will have it sent to their address they put. The DAO will get a fee in gOHM every time the swap to DAI happens. This will essentially pay for the gas costs of up-keeping the contract.

Users also have the option to withdraw at any time or claim their yield at any time in gOhm or DAI.

Not Done:
Have no tested on testnet will do that after writing tests and looks good locally.

### DIFF
--- a/contracts/peripheral/YieldStreamer.sol
+++ b/contracts/peripheral/YieldStreamer.sol
@@ -44,10 +44,10 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
 
     struct RecipientInfo {
         address recipientAddress;
-        uint256 lastUpkeepTimestamp;
-        uint256 paymentInterval; // Time before yield is able to be swapped to stream tokens
-        uint256 unclaimedStreamTokens;
-        uint256 userMinimumAmountThreshold;
+        uint128 lastUpkeepTimestamp;
+        uint128 paymentInterval; // Time before yield is able to be swapped to stream tokens
+        uint128 unclaimedStreamTokens;
+        uint128 userMinimumAmountThreshold;
     }
 
     mapping(uint256 => RecipientInfo) public recipientInfo; // depositId -> RecipientInfo
@@ -81,8 +81,8 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
         address sushiRouter_,
         address staking_,
         address authority_,
-        uint256 maxSwapSlippagePercent_,
-        uint256 feeToDaoPercent_,
+        uint128 maxSwapSlippagePercent_,
+        uint128 feeToDaoPercent_,
         uint256 minimumTokenThreshold_
     ) YieldSplitter(sOHM_) OlympusAccessControlled(IOlympusAuthority(authority_)) {
         gOHM = gOHM_;
@@ -110,8 +110,8 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
     function deposit(
         uint256 amount_,
         address recipient_,
-        uint256 paymentInterval_,
-        uint256 userMinimumAmountThreshold_
+        uint128 paymentInterval_,
+        uint128 userMinimumAmountThreshold_
     ) external override {
         if (depositDisabled) revert YieldStreamer_DepositDisabled();
         if (amount_ <= 0) revert YieldStreamer_InvalidAmount();
@@ -121,7 +121,7 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
 
         recipientInfo[depositId] = RecipientInfo({
             recipientAddress: recipient_,
-            lastUpkeepTimestamp: block.timestamp,
+            lastUpkeepTimestamp: uint128(block.timestamp),
             paymentInterval: paymentInterval_,
             unclaimedStreamTokens: 0,
             userMinimumAmountThreshold: userMinimumAmountThreshold_
@@ -207,7 +207,7 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
         if (withdrawDisabled) revert YieldStreamer_WithdrawDisabled();
         if (recipientInfo[id_].recipientAddress != msg.sender) revert YieldStreamer_UnauthorisedAction();
 
-        recipientInfo[id_].lastUpkeepTimestamp = block.timestamp;
+        recipientInfo[id_].lastUpkeepTimestamp = uint128(block.timestamp);
 
         uint256 yield = _redeemYield(id_);
 
@@ -222,7 +222,7 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
         if (withdrawDisabled) revert YieldStreamer_WithdrawDisabled();
         if (recipientInfo[id_].recipientAddress != msg.sender) revert YieldStreamer_UnauthorisedAction();
 
-        recipientInfo[id_].lastUpkeepTimestamp = block.timestamp;
+        recipientInfo[id_].lastUpkeepTimestamp = uint128(block.timestamp);
 
         uint256 gOHMYield = _redeemYield(id_);
         uint256 totalOhmToSwap = staking.unwrap(address(this), gOHMYield);
@@ -260,7 +260,7 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
         @param id_ Id of the deposit
         @param threshold_ amount of streamTokens
     */
-    function updateUserMinDaiThreshold(uint256 id_, uint256 threshold_) external override {
+    function updateUserMinDaiThreshold(uint256 id_, uint128 threshold_) external override {
         if (threshold_ < minimumTokenThreshold) revert YieldStreamer_MinTokenThresholdTooLow();
         if (depositInfo[id_].depositor != msg.sender) revert YieldStreamer_UnauthorisedAction();
 
@@ -272,7 +272,7 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
         @param id_ Id of the deposit
         @param paymentInterval_ amount of time in seconds
     */
-    function updatePaymentInterval(uint256 id_, uint256 paymentInterval_) external override {
+    function updatePaymentInterval(uint256 id_, uint128 paymentInterval_) external override {
         if (depositInfo[id_].depositor != msg.sender) revert YieldStreamer_UnauthorisedAction();
 
         recipientInfo[id_].paymentInterval = paymentInterval_;
@@ -319,8 +319,10 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
             if (_isUpkeepEligible(currentId)) {
                 RecipientInfo storage currentrecipientInfo = recipientInfo[currentId];
 
-                currentrecipientInfo.lastUpkeepTimestamp = block.timestamp;
-                currentrecipientInfo.unclaimedStreamTokens += (amounts[1] * _redeemYield(currentId)) / totalGOHM;
+                currentrecipientInfo.lastUpkeepTimestamp = uint128(block.timestamp);
+                currentrecipientInfo.unclaimedStreamTokens += uint128(
+                    (amounts[1] * _redeemYield(currentId)) / totalGOHM
+                );
 
                 if (currentrecipientInfo.unclaimedStreamTokens >= currentrecipientInfo.userMinimumAmountThreshold) {
                     uint256 streamTokensToSend = currentrecipientInfo.unclaimedStreamTokens;
--- a/contracts/peripheral/YieldStreamer.sol
+++ b/contracts/peripheral/YieldStreamer.sol
@@ -117,8 +117,6 @@ contract YieldStreamer is IYieldStreamer, YieldSplitter, OlympusAccessControlled
         if (amount_ <= 0) revert InvalidAmount();
         if (userMinimumDaiThreshold_ < minimumDaiThreshold) revert MinDaiThresholdTooLow();
 
-        IERC20(gOHM).safeTransferFrom(msg.sender, address(this), amount_);
-
         uint256 depositId = _deposit(msg.sender, amount_);
 
         recipientInfo[depositId] = RecipientInfo({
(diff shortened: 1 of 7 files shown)


ind-igo:
I thought this would be using Tyche as a baseline. If not, should we instead systematize how this works? I hesitate on making two contracts that split yield. Would rather have 1 that is used in multiple ways.

ind-igo (changes requested):

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,466 @@
...
+        @param id_ Id of the deposit.
+        @param amount_ Amount of sOHM to withdraw.
+    */
+    function withdrawPrincipal(uint256 id_, uint256 amount_) external override {
+        require(!withdrawDisabled, "Withdraws currently disabled");
+        require(amount_ > 0, "Invalid withdraw amount");
+
+        DepositInfo storage userDeposit = depositInfo[id_];
+        require(userDeposit.depositor == msg.sender, "Deposit is not yours");
+        require(amount_ < userDeposit.principalAmount, "input >= principal");
```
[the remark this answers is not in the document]
Paul4912: Im doing this not just for that but so amount cant equal principalAmount. If user wants to withdraw all their principal I want them to use withdrawAll() so that internally I can do the cleanup of their deposit from the activeDepositIds. So i was intending this method only for partial withdrawals of principal.

I could allow them to withdraw it all here and leave the yield and unclaimed dai in but that doesn't seem like they would want to do that intentionally.

Im thinking now to do a if statement if amount = principal then just call withdrawAll() instead. They might not expect withdrawing their yield as well and unclaimed dai but that seems better for the user since they would want to do that anyway.

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,466 @@
...
+        uint256 feeToDao = (totalOhm * feeToDaoPercent) / 1000;
+        IERC20(sOHM).safeTransfer(authority.governor(), feeToDao);
+        uint256 totalOhmToSwap = totalOhm - feeToDao;
+        staking.unstake(address(this), totalOhmToSwap, false, false);
+
+        require(IERC20(OHM).approve(address(sushiRouter), totalOhmToSwap), "approve failed");
+        uint256[] memory calculatedAmounts = sushiRouter.getAmountsOut(totalOhmToSwap, sushiRouterPath);
+        uint256[] memory amounts = sushiRouter.swapExactTokensForTokens(
+            totalOhmToSwap,
+            (calculatedAmounts[1] * (1000 - maxSwapSlippagePercent)) / 1000,
```
tonypiano, line 326: Do you think the way we are calculating the max slippage here could leave this open to a sandwich attack? Take a look at the Uniswap V2 [docs](https://docs.uniswap.org/protocol/V2/guides/smart-contract-integration/trading-from-a-smart-contract)

> When trading from a smart contract, the most important thing to keep in mind is that access to an external price source is required. Without this, trades can be frontrun for considerable loss.

Maybe something like this:
* Flashloan a bunch of ETH
* Buy DAI via ETH->OHM-DAI, pushing the price of OHM down
* run `upkeep`, selling OHM for a small amount of DAI
* buyback ETH via DAI->OHM-ETH claiming the amount of DAI that should have been swapped for here.

My guess is the OHM-DAI pool has too much liquidity and these amounts will be too small to make this profitable.

tonypiano (commented):
Small comment on the swap, but the rest looks pretty good.

contracts/peripheral/YieldStreamer.sol
Paul4912: Yea so from our discussions either the amount would be too small for MEV attacks. Or if that is not the case route the transaction through flashbots rpc. Haven't looked into that yet just took Indigos word for it. Just assumed there would be a way to avoid getting attacked so i've put it on the backfoot for now.

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,447 @@
...
+        if (depositInfo[id_].depositor != msg.sender && recipientInfo[id_].recipientAddress != msg.sender)
+            revert UnauthorisedAction();
+
+        recipientInfo[id_].lastUpkeepTimestamp = block.timestamp;
+
+        uint256 gOHMYield = _redeemYield(id_);
+        uint256 totalOhmToSwap = staking.unwrap(address(this), gOHMYield);
+        staking.unstake(address(this), totalOhmToSwap, false, false);
+
+        IERC20(OHM).approve(address(sushiRouter), totalOhmToSwap);
```
kushal256, line 234: Use SafeERC20::approve?  Also might want to consider doing this in the constructor for large amount (0xFFFFFF) to save gas so doesn't need to be performed each time

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,447 @@
...
+    /************************
+     * Restricted Setter Functions
+     ************************/
+
+    /**
+        @notice Setter for maxSwapSlippagePercent.
+        @param slippagePercent_ new slippage value as a fraction / 1000.
+    */
+    function setMaxSwapSlippagePercent(uint256 slippagePercent_) external onlyGovernor {
+        maxSwapSlippagePercent = slippagePercent_;
```
kushal256, line 406: add a sanity check to ensure < 1000?

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,447 @@
...
+        IERC20(DAI).safeTransfer(msg.sender, daiToSend);
+    }
+
+    /**
+        @notice harvest all your unclaimed Dai
+        @param id_ Id of the deposit
+    */
+    function harvestDai(uint256 id_) external override {
+        if (withdrawDisabled) revert WithdrawDisabled();
+        if (depositInfo[id_].depositor != msg.sender && recipientInfo[id_].recipientAddress != msg.sender)
```
kushal256, line 255: I thought the point of `recipientAddress` is so yield can be sent elsewhere?  Am I misreading this or does this enforce `recipientAddress` is the same as depositor?

kushal256 (commented):
Overall very nice!

Also had a nitpick question, why are all variable names for function paramters ending with '_'?  Not really sure what the desired standard is, but this seems different than what's in `main` right now.

contracts/peripheral/YieldStreamer.sol
Paul4912: Yea i shoulda done just recipient.

ind-igo (changes requested):

Paul4912:
> 

not sure who started it but i was just following the convention set in place. Kinda like it makes it clearer which variable is parameter.

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,447 @@
...
+
+    bool public depositDisabled;
+    bool public withdrawDisabled;
+    bool public upkeepDisabled;
+
+    uint256 public maxSwapSlippagePercent; // as fraction / 1000
+    uint256 public feeToDaoPercent; // as fraction /1000
+    uint256 public minimumDaiThreshold;
+
+    struct RecipientInfo {
```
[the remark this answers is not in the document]
Paul4912: yea i agree.
im also thinking userMinimumDaiThreshold and unclaimed dai might be able to be uint128. Thats still 38 decimals can fit up to 10^20 DAI. what do you think.

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,442 @@
...
+        @param paymentInterval_ amount of time in seconds
+    */
+    function updatePaymentInterval(uint256 id_, uint256 paymentInterval_) external override {
+        if (depositInfo[id_].depositor != msg.sender) revert YieldStreamer_UnauthorisedAction();
+
+        recipientInfo[id_].paymentInterval = paymentInterval_;
+    }
+
+    /**
+        @notice Upkeeps all deposits if they are eligible to be upkept. Converts excess yield from gOHM to streamTokens.
```
ind-igo, line 282: What does upkeep mean in this context? Could you clarify this comment?

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,442 @@
...
+    /**
+        @notice Get deposits principal amount in gOhm
+        @param id_ Id of the deposit.
+     */
+    function getPrincipalInGOHM(uint256 id_) external view returns (uint256) {
+        return IgOHM(gOHM).balanceTo(depositInfo[id_].principalAmount);
+    }
+
+    /**
+        @notice Returns whether deposit id is eligible for upkeep
```
ind-igo, line 380: Please clarify somewhere what Upkeep means. It is not very clear to me.

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,442 @@
...
+        @return bool
+     */
+    function _isUpkeepEligible(uint256 id_) internal view returns (bool) {
+        if (block.timestamp >= recipientInfo[id_].lastUpkeepTimestamp + recipientInfo[id_].paymentInterval) {
```
ind-igo, line 384: can just return this statement and save 3 lines.

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,442 @@
...
+        @param slippagePercent_ new slippage value as a fraction / 1000.
+    */
+    function setMaxSwapSlippagePercent(uint256 slippagePercent_) external onlyGovernor {
+        if (slippagePercent_ > 1000000) revert YieldStreamer_InvalidAmount();
+        maxSwapSlippagePercent = slippagePercent_;
+    }
+
+    /**
+        @notice Setter for feeToDaoPercent.
+        @param feePercent_ new fee value as a fraction / 1000.
```
ind-igo, line 405: would add `(4 decimals)` at the end of this comment for clarity.

ind-igo: or just say `1e4`

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,442 @@
...
+    */
+    function setFeeToDaoPercent(uint256 feePercent_) external onlyGovernor {
+        if (feePercent_ > 1000000) revert YieldStreamer_InvalidAmount();
```
ind-igo, line 408: would replace `1000000` with `1e7`. Much easier to parse, since you don't need to count the zeroes.

ind-igo (changes requested):

contracts/peripheral/YieldStreamer.sol
Paul4912: comment is old i made it 6 decimals since that seems to be standard used in alot of other contracts i see

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,447 @@
...
+                // Delete integer from array by swapping with last element and calling pop()
+                if (recipientIdsArray[i] == id_) {
+                    recipientIdsArray[i] = recipientIdsArray[recipientIdsArray.length - 1];
+                    recipientIdsArray.pop();
+                    break;
+                }
+            }
+
+            IERC20(gOHM).safeTransfer(msg.sender, principal);
+            IERC20(gOHM).safeTransfer(recipient, totalGOHM - principal);
```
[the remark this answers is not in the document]
Paul4912: principal here is returned from closeDeposit is returned as agnostic. just internally calculated as sOhm. Try to stick to the rule of internally calculate principal in sOhm and externally return everything in agnostic gOhm.

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,447 @@
...
+    function withdrawYield(uint256 id_) external override {
+        if (withdrawDisabled) revert WithdrawDisabled();
+        if (depositInfo[id_].depositor != msg.sender && recipientInfo[id_].recipientAddress != msg.sender)
+            revert UnauthorisedAction();
+
+        recipientInfo[id_].lastUpkeepTimestamp = block.timestamp;
+
+        uint256 yield = _redeemYield(id_);
+
+        IERC20(gOHM).safeTransfer(msg.sender, yield);
```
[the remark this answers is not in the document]
Paul4912: yea thats probably safer.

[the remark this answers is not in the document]
Paul4912: what about withdraw principal. would you send it to depositor instead of msg.sender even though we have a check for depositor == msg.sender?

contracts/peripheral/YieldStreamer.sol
```diff
@@ -0,0 +1,447 @@
...
+        @param id_ Id of the deposit
+    */
+    function withdrawYieldAsDai(uint256 id_) external override {
+        if (withdrawDisabled) revert WithdrawDisabled();
+        if (depositInfo[id_].depositor != msg.sender && recipientInfo[id_].recipientAddress != msg.sender)
+            revert UnauthorisedAction();
+
+        recipientInfo[id_].lastUpkeepTimestamp = block.timestamp;
+
+        uint256 gOHMYield = _redeemYield(id_);
```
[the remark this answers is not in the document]
Paul4912: i'l stick to gOHM since original gOHM contract is like that.

PR #189, opened 2022-01-22
Update LiquidityMigrator.sol

olyzeus:


contracts/migration/LiquidityMigrator.sol
```diff
@@ -2,92 +2,110 @@
...
+        
+        // deposit the paired token into the new treasury
+        uint256 amountToken = token.balanceOf(address(this));
+
+        // burn the old OHM
+        uint256 amountOld = oldOHM.balanceOf(address(this));
+        oldOHM.burn(amountOld);
+
+        // mint new OHM after converting to new supply
+        uint256 amountNew = amountOld * newsohm.index() / oldsohm.index();
```
lolchocotaco, line 94: Instead of having this contract mint new ohm, we accomplished this by calling the migrator. Keeps the logic / bookkeeping all in one place. What's the reasoning for doing the accounting in this contract?

contracts/migration/LiquidityMigrator.sol
```diff
@@ -2,92 +2,110 @@
...
+        newOHM.approve(address(router), amountNew);
+        token.approve(address(router), amountToken);
+
+        router.addLiquidity(
+            address(token), 
+            address(newOHM), 
+            amountToken, 
+            amountNew, 
+            amountToken * 99 / 100, 
+            amountNew * 99 / 100,
```
lolchocotaco, line 106: Why multiply by 99%?

contracts/migration/LiquidityMigrator.sol
```diff
@@ -2,92 +2,110 @@
...
-     */
-    function migrateLP(
-        address pair,
-        IUniswapV2Router routerFrom,
-        IUniswapV2Router routerTo,
-        address token,
-        uint256 _minA,
-        uint256 _minB,
-        uint256 _deadline
+    function remove(
```
lolchocotaco, line 25: In which scenarios do we need this? 
Wouldn't we want to migrate only?

lolchocotaco (commented):
I have some questions on separation of concern and "why"

contracts/migration/LiquidityMigrator.sol
```diff
@@ -2,92 +2,110 @@
 pragma solidity 0.7.5;
 
 import "../interfaces/IERC20.sol";
-import "../libraries/SafeERC20.sol";
-import "../libraries/SafeMath.sol";
```
ind-igo, line 6: This is necessary

contracts/migration/LiquidityMigrator.sol
```diff
@@ -2,92 +2,110 @@
...
+        uint256 minOhm,
+        uint256 deadline
     ) external onlyGovernor {
-        // Since we are adding liquidity, any existing balance should be excluded
-        uint256 initialNewOHMBalance = newOHM.balanceOf(address(this));
         // Fetch the treasury balance of the given liquidity pair
-        uint256 oldLPAmount = IERC20(pair).balanceOf(address(oldTreasury));
-        oldTreasury.manage(pair, oldLPAmount);
+        uint256 max = pair.balanceOf(address(oldTreasury));
+        uint256 lpAmount = amount < max ? amount : max;
```
ind-igo, line 36: need safemath

contracts/migration/LiquidityMigrator.sol
ind-igo: Also needs safemath

contracts/migration/LiquidityMigrator.sol
ind-igo: Also needs safemath

ind-igo (changes requested):

contracts/migration/LiquidityMigrator.sol
olyzeus: the migrator does not have corresponding gOHM for the tokens in this pool afaik. verified at the time of writing but can be checked with: gOHM in migrator * index vs (sOHMv1 circulating * 46.7) + wsOHM circulating + OHMv1 circulating. intent is to avoid a deficit in that pool, as staked supply was used for the number of gOHM to mint at migration.

contracts/migration/LiquidityMigrator.sol
olyzeus: min amounts could be passed in, but really all we need to ensure is it has not skewed significantly from the balance on the v1 side.

contracts/migration/LiquidityMigrator.sol
olyzeus: i had considered proposing the removal of this liquidity rather than migration, since LPs currently make up an unprecedented % of total supply. if said proposal does not occur then yes, unnecessary

contracts/migration/LiquidityMigrator.sol
olyzeus: ok dad

contracts/migration/LiquidityMigrator.sol
olyzeus: we get it, just push the change ser

PARABRAHMAN0:
Seeeems to be stale. Closing.

PR #199, opened 2022-02-08
Tyche gOHM

0xLienid:
Create new Yield Director contract that works with the new YieldSplitter abstract and accepts gOHM deposits.


contracts/peripheral/YieldDirectorGohm.sol
```diff
@@ -0,0 +1,363 @@
...
+import {YieldSplitter} from "../types/YieldSplitter.sol";
+import {OlympusAccessControlled, IOlympusAuthority} from "../types/OlympusAccessControlled.sol";
+
+/**
+    @title YieldDirector (codename Tyche) 
+    @notice This contract allows donors to deposit their sOHM and donate their rebases
+            to any address. Donors will be able to withdraw their principal
+            sOHM at any time. Donation recipients can also redeem accrued rebases at any time.
+ */
+contract YieldDirectorGohm is YieldSplitter, OlympusAccessControlled {
```
Paul4912, line 18: Were u meant to implement IYieldDirectory here?

contracts/interfaces/IYieldDirector.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity >=0.7.5;
+
+interface IYieldDirector {
```
Paul4912, line 4: Isn't this file already exist in main branch for tyche v1?

contracts/peripheral/YieldDirectorGohm.sol
```diff
@@ -0,0 +1,363 @@
...
+import {IsOHM} from "../interfaces/IsOHM.sol";
+import {IgOHM} from "../interfaces/IgOHM.sol";
+import {IStaking} from "../interfaces/IStaking.sol";
+import {SafeERC20} from "../libraries/SafeERC20.sol";
+import {YieldSplitter} from "../types/YieldSplitter.sol";
+import {OlympusAccessControlled, IOlympusAuthority} from "../types/OlympusAccessControlled.sol";
+
+/**
+    @title YieldDirector (codename Tyche) 
+    @notice This contract allows donors to deposit their sOHM and donate their rebases
```
Paul4912, line 14: comment still says sOhm

contracts/peripheral/YieldDirectorGohm.sol
Paul4912, line 18: What do you think of yielddirectorv2 naming scheme instead of this?

contracts/peripheral/YieldDirectorGohm.sol
```diff
@@ -0,0 +1,363 @@
...
+        require(staking_ != address(0), "Invalid address for staking");
+
+        sOHM = sOhm_;
+        staking = IStaking(staking_);
+    }
+
+    /************************
+    * Modifiers
+    ************************/
+    modifier isValidDeposit(uint256 amount_, address recipient_) {
```
Paul4912, line 57: I think were trying to move away from modifiers. While they provide nice visual benefit, under the hood it seems like are not more efficient than internal functions in reducing code size due to how they're implemented. 

I think internal function returning true or false should do the same job more efficiently.

Heres someone raising that issue up. 
https://github.com/ethereum/solidity/issues/6584

contracts/peripheral/YieldDirectorGohm.sol
```diff
@@ -0,0 +1,363 @@
...
+        }
+
+        return agnosticRedeemable;
+    }
+
+    /**
+        @notice Redeem recipient's donated amount of sOHM at current index from one donor
+        @param id_ Deposit id for this donation
+     */
+    function redeemYield(uint256 id_) external returns (uint256) {
```
Paul4912, line 316: remember seeing ur message about closing deposit after redeem yield when principal is 0 instead of on withdraw. Is that not done yet?

contracts/peripheral/YieldDirectorGohm.sol
0xLienid: Like that much better!

contracts/peripheral/YieldDirectorGohm.sol
0xLienid: Cool, will adjust this

PR #200, opened 2022-02-08
Deployed contract addresses saved to file

0xJem:
This PR throws out the previous approach of storing the deployment files (any time a contract was updated, it would invalidate the old file and require a manual update of the constants in olympus-frontend).

Instead, we now write the addresses into a JSON file in `addresses/addresses.json`. This can be used by a local dev, but can also be passed to the frontend and ingested dynamically. Magic!


hardhat.config.ts
```diff
@@ -52,6 +52,12 @@ const config: HardhatUserConfig = {
             forking: {
                 url: `https://eth-mainnet.alchemyapi.io/v2/${alchemyApiKey}`,
             },
+            accounts: {
+                // This will result in 10000 ETH being deposited into this wallet on the local testnet
+                // Obviously don't use this on mainnet
+                mnemonic:
+                    "subway bright wet favorite tide muscle alley marble must ecology result treat",
```
ind-igo, line 59: ser

ind-igo (changes requested):

hardhat.config.ts
0xJem: It's a completely random wallet - not mine.

0xJem: We need to enable the developer to have ETH deposited into their wallet. Is there a different approach you'd suggest? Let the dev provide the seed phrase as an env variable and have that imported into this?

ind-igo: yup, we have a .env.example that has a template for a .env file. if you want to work off of that. Can the method for getting ETH deposited into the wallet work off of Private keys?

0xJem: Have removed the commits. Can you delete the comment so the preview also disappears?

PARABRAHMAN0:
We really need a sitdown on this topic. I would leave this open simply for this reason. I've been writing mini util with importable addresses in /test/utils on the allocator branch, and was thinking about directly writing. I don't see why we should rely on something like `hardhat-deploy` to do stuff for us when we can just code it ourselves.

philipjonsen:
Yes, smart contracts are the root of all DAO projects. But integrating and and working could with them could be better, before any broader applications this should be checked one, five or 10-100 times by Security professionals.

PR #208, opened 2022-02-11
YieldDirector + YieldSplitter without recipient

0xLienid:
Updated versions of YieldDirectorV2 and YieldSplitter per discussions with Paul and Indigo about moving the recipient tracking up to the inheriting contract level for more modularity of YieldSplitter


contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+    IStaking public immutable staking;
+    mapping(address => uint256[]) public recipientIds;
+    mapping(uint256 => address) internal recipientLookup;
+
+    bool public depositDisabled;
+    bool public withdrawDisabled;
+    bool public redeemDisabled;
+
+    event Deposited(address indexed donor_, address indexed recipient_, uint256 amount_);
+    event DepositUpdated(address indexed donor_, address indexed recipient_, uint256 amount_);
```
Paul4912, line 31: Why not have both deposit and depositUpdate as same event? They are both essentially a user adding money into a deposit im not seeing the need for why they need seperate events. I feel like update implies some configuration or parameter change when its just adding more money in.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+        @param id_ Deposit ID to remove gOHM debt from
+        @param amount_ Amount of gOHM debt to remove and return to donor
+     */
+    function withdrawPrincipal(uint256 id_, uint256 amount_) external {
+        require(isValidWithdrawal(amount_), "Not a valid withdrawal.");
+        DepositInfo storage currDeposit = depositInfo[id_];
+        require(currDeposit.depositor == msg.sender, "Can't withdraw principal that's not yours");
+
+        if (amount_ >= IgOHM(gOHM).balanceTo(currDeposit.principalAmount)) {
+            _withdrawPrincipal(id_, amount_);
```
Paul4912, line 150: this will revert right? 
here were meant to make amount_ = IgOHM(gOHM).balanceTo(currDeposit.principalAmount)
or set a new variable called amountToWithdraw = IgOHM(gOHM).balanceTo(currDeposit.principalAmount).

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+        @notice Withdraw donor's gOHM from vault, and return it as sOHM
+        @param id_ Deposit ID to remove gOHM debt from
+        @param amount_ Amount of gOHM debt to remove and return to donor as sOHM
+     */
+    function withdrawPrincipalAsSohm(uint256 id_, uint256 amount_) external {
+        require(isValidWithdrawal(amount_), "Not a valid withdrawal.");
+        DepositInfo storage currDeposit = depositInfo[id_];
+        require(currDeposit.depositor == msg.sender, "Can't withdraw principal that's not yours");
+
+        if (amount_ >= IgOHM(gOHM).balanceTo(currDeposit.principalAmount)) {
```
Paul4912, line 175: if withdraw sOhm we can directly compare with currDeposit.principalAmount.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+
+        uint256 depositsLength = depositIds.length;
+        require(depositsLength != 0, "User not donating to anything");
+
+        uint256 principalTotal = 0;
+
+        for (uint256 index = 0; index < depositsLength; index++) {
+            DepositInfo storage currDeposit = depositInfo[depositIds[index]];
+            principalTotal += currDeposit.principalAmount;
+            _withdrawPrincipal(currDeposit.id, IgOHM(gOHM).balanceTo(currDeposit.principalAmount));
```
Paul4912, line 209: would it be helpful if base contract had a function _withdrawAllprincipal(uint256 id). Thats withdraws all the principal and returns the amount.

Seems reasonably generic to add to base contract to me. What u think.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+        @param amount_ Amount of gOHM debt to remove and return to donor as sOHM
+     */
+    function withdrawPrincipalAsSohm(uint256 id_, uint256 amount_) external {
+        require(isValidWithdrawal(amount_), "Not a valid withdrawal.");
+        DepositInfo storage currDeposit = depositInfo[id_];
+        require(currDeposit.depositor == msg.sender, "Can't withdraw principal that's not yours");
+
+        if (amount_ >= IgOHM(gOHM).balanceTo(currDeposit.principalAmount)) {
+            _withdrawPrincipal(id_, amount_);
+            currDeposit.principalAmount = 0;
```
Paul4912, line 177: withdraw principal takes care of this right

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    /**
+        @notice Return arrays of donor's recipients and deposit amounts (gOHM value based on sOHM equivalent deposit), matched by index
+        @param donor_ Address of user donating yield
+     */
+    function getAllDeposits(address donor_) external view returns (address[] memory, uint256[] memory) {
+        uint256[] storage depositIds = depositorIds[donor_];
+
+        uint256 len = depositIds.length == 0 ? 1 : depositIds.length;
```
Paul4912, line 268: didn't realise we could use these conditiona operators in solidity.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+        @notice Return total amount of gOHM donated to recipient since last full withdrawal or redemption
+        @param donor_ Address of user donating yield
+        @param recipient_ Address of user recieiving donated yield
+     */
+    function donatedTo(address donor_, address recipient_) external view returns (uint256) {
+        uint256[] storage depositIds = depositorIds[donor_];
+
+        for (uint256 index = 0; index < depositIds.length; index++) {
+            if (recipientLookup[depositIds[index]] == recipient_) {
+                return _getOutstandingYield(depositInfo[depositIds[index]].principalAmount, depositInfo[depositIds[index]].agnosticAmount);
```
Paul4912, line 297: wouldn't it be simpler to use this function you made
redeemableBalance(uint256 id_) and just chuck in the id. 
makes code look shorter and less chance of error putting in wrong value into _getOutstandingYield()

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+    function redeemYieldAsSohm(uint256 id_) external returns (uint256) {
+        require(!redeemDisabled, "Redeems currently disabled");
+        require(recipientLookup[id_] == msg.sender);
+
+        uint256 amountRedeemed = _redeemYield(id_);
+        require(amountRedeemed > 0, "No redeemable balance");
+
+        IERC20(sOHM).approve(address(staking), amountRedeemed);
+        staking.unwrap(msg.sender, amountRedeemed);
+
```
Paul4912, line 377: forgot to send sOHM to user?

ind-igo (changes requested):

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,145 @@
...
+    function _deposit(
+        address depositor_,
+        uint256 amount_
+    ) internal returns (uint256 depositId) {
+        depositorIds[depositor_].push(idCount);
+
+        depositInfo[idCount] = DepositInfo({
+            id: idCount,
+            depositor: depositor_,
+            principalAmount: IgOHM(gOHM).balanceFrom(amount_),
```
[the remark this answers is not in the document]
Paul4912: we could easily reimplement those functions and use an internal function. 

The internal function would use sOhm to get the index. So for other chains we would replace that with an oracle then.

contracts/peripheral/YieldDirectorV2.sol
0xLienid: doesn't revert, but probably cleaner to do it your way and not in-line. will change

contracts/peripheral/YieldDirectorV2.sol
0xLienid: This is withdrawing the passed amount of gOHM as sOHM so have to compare the amount passed in to the gOHM equivalent of the principal amount. Could change it so that the value passed in is sOHM. What does everyone think?

contracts/peripheral/YieldDirectorV2.sol
0xLienid: Yeah think that would be useful and generic enough to add

contracts/peripheral/YieldDirectorV2.sol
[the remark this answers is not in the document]
0xLienid: doesn't compile doing either return (address(0), 0) or return ([address(0)], [0]). Those aren't automatically convertible to the return type apparently

contracts/peripheral/YieldDirectorV2.sol
0xLienid: yep, probably

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+            to any address. Donors will be able to withdraw the sOHM equivalent of their principal
+            gOHM at any time. Donation recipients can also redeem accrued rebases at any time.
+ */
+contract YieldDirectorV2 is YieldSplitter, OlympusAccessControlled {
+    using SafeERC20 for IERC20;
+
+    address public immutable sOHM;
+    IStaking public immutable staking;
+    mapping(address => uint256[]) public recipientIds;
+    mapping(uint256 => address) internal recipientLookup;
```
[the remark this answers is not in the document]
0xLienid: done

contracts/peripheral/YieldDirectorV2.sol
0xLienid: done

contracts/peripheral/YieldDirectorV2.sol
0xLienid: done

contracts/peripheral/YieldDirectorV2.sol
0xLienid: done

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,441 @@
...
+
+        sOHM = sOhm_;
+        staking = IStaking(staking_);
+    }
+
+    /************************
+     * Modifiers
+     ************************/
+    function isValidDeposit(uint256 amount_, address recipient_) internal returns (bool) {
+        return !(depositDisabled || amount_ <= 0 || recipient_ == address(0));
```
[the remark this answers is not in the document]
0xLienid: done for all errors with strings

contracts/peripheral/YieldDirectorV2.sol
Paul4912: I think im going to make a pr with yieldsplitter as only file change that way we can both rebase on it without having to coodinate with each other on all the changes,

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,451 @@
...
+    */
+    function redeemYield(uint256 id_) external returns (uint256) {
+        if (redeemDisabled) revert YieldDirector_RedeemsDisabled();
+        require(recipientLookup[id_] == msg.sender);
+
+        uint256 amountRedeemed = _redeemYield(id_);
+        if (amountRedeemed == 0) revert YieldDirector_NoRedeemableBalance();
+
+        IERC20(gOHM).safeTransfer(msg.sender, amountRedeemed);
+
```
Paul4912, line 370: Is here where you want to add the logic:
if principal == 0. remove deposit from recipient array and deposit to recipient lookup ?

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,451 @@
...
+    function redeemAllYield() external returns (uint256) {
+        if (redeemDisabled) revert YieldDirector_RedeemsDisabled();
+
+        uint256 totalRedeemable = totalRedeemableBalance(msg.sender);
+        if (totalRedeemable == 0) revert YieldDirector_NoRedeemableBalance();
+
+        uint256[] storage receiptIds = recipientIds[msg.sender];
+        
+        for (uint256 index = 0; index < receiptIds.length; index++) {
+            _redeemYield(receiptIds[index]);
```
Paul4912, line 403: Since you are already looping through the recieiptIds here i think you can do the adding up totalredeemable yield here. 

before the for loop
uint256 totalRedeemable = 0
inside for loop
totalRedeemable += _redeemYield(receiptIds[index]);

This way we don't have to call totalRedeemableBalance which loops through that array again.

contracts/peripheral/YieldDirectorV2.sol
Paul4912: Also can we move this piece of code:
if (userDeposit.principalAmount == 0) {
            _closeDeposit(id_);
        }
 From base contract to implementation. I think it might cause some issues for some implementations. I think since implementation handles the recipient logic they should probably also handle closing deposits when they need to.

contracts/peripheral/YieldDirectorV2.sol
0xLienid: oh. sorry, misinterpreted that. Yes, it will revert if the user tries to withdraw the max amount I believe. I'll test now. I think that probably having it instead default to withdrawing the max amount if they request more is the better way to go about it. Thoughts @ind-igo?

contracts/peripheral/YieldDirectorV2.sol
0xLienid: yes think so, and good call. I'll do that.

contracts/peripheral/YieldDirectorV2.sol
0xLienid: good call

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,453 @@
...
+    error YieldDirector_NotYourDeposit();
+    error YieldDirector_NoDeposits();
+    error YieldDirector_NoRedeemableBalance();
+    error YieldDirector_WithdrawalsDisabled();
+    error YieldDirector_RedeemsDisabled();
+
+    address public immutable sOHM;
+    IStaking public immutable staking;
+    mapping(address => uint256[]) public recipientIds;
+    mapping(uint256 => address) public recipientLookup;
```
Paul4912, line 37: For future readers probably good idea to put comment like depositId -> recipient address so they know what uint256 is here.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,453 @@
...
+    error YieldDirector_InvalidWithdrawal();
+    error YieldDirector_NotYourDeposit();
+    error YieldDirector_NoDeposits();
+    error YieldDirector_NoRedeemableBalance();
+    error YieldDirector_WithdrawalsDisabled();
+    error YieldDirector_RedeemsDisabled();
+
+    address public immutable sOHM;
+    IStaking public immutable staking;
+    mapping(address => uint256[]) public recipientIds;
```
Paul4912, line 36: similar here, comment like this would be nice for out of context readers.
address -> Array of the deposit id's user is recipient of

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+        uint256[] memory receiptIds = recipientIds[msg.sender];
+        
+        for (uint256 index = 0; index < receiptIds.length; ++index) {
+            uint256 currRedemption = _redeem(receiptIds[index]);
+            amountRedeemed += currRedemption;
+
+            emit Donated(depositInfo[receiptIds[index]].depositor, msg.sender, currRedemption);
+        }
+
+        if (amountRedeemed == 0) revert YieldDirector_NoRedeemableBalance();
```
Paul4912, line 464: this seems kinda redundant because above you call _redeem and inside redeem() if it redeems nothing then it reverts anyway for one deposit. So this will never be 0.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+
+        emit Deposited(msg.sender, recipient_, amount_);
+    }
+
+    /**
+        @notice Increases the amount of gOHM directing yield to a recipient
+        @param depositId_ The global ID number of the deposit to add the additional deposit to
+        @param amount_ Quantity of new gOHM deposited redirecting yield to the current deposit's recipient
+    */
+    function _increaseDeposit(uint256 depositId_, uint256 amount_) internal {
```
Paul4912, line 388: is this function really necessary if its just these 2 lines?

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+        @param depositId_ Deposit ID to remove gOHM deposit from
+        @param amount_ Amount of gOHM deposit to remove and return to donor 
+    */
+    function _withdraw(uint256 depositId_, uint256 amount_) internal returns (uint256 amountWithdrawn) {
+        if (isInvalidWithdrawal(amount_)) revert YieldDirector_InvalidWithdrawal();
+
+        DepositInfo storage currDeposit = depositInfo[depositId_];
+        uint256 withdrawableBalance = _toAgnostic(currDeposit.principalAmount);
+
+        if (amount_ >= withdrawableBalance) {
```
Paul4912, line 405: how bad would it be to just instead of making withdrawableBalance by storing 2 variables just go 
depositInfo[depositId_]. principalAmount ?

Seems like that might save some gas since those 2 variables are only used once.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+        emit Deposited(depositInfo[depositId_].depositor, recipientLookup[depositId_], amount_);
+    }
+
+    /**
+        @notice Withdraw gOHM deposit from vault
+        @param depositId_ Deposit ID to remove gOHM deposit from
+        @param amount_ Amount of gOHM deposit to remove and return to donor 
+    */
+    function _withdraw(uint256 depositId_, uint256 amount_) internal returns (uint256 amountWithdrawn) {
+        if (isInvalidWithdrawal(amount_)) revert YieldDirector_InvalidWithdrawal();
```
Paul4912, line 400: missing 
if (withdrawDisabled) revert YieldDirector_WithdrawalsDisabled(); ?

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+        staking.unwrap(msg.sender, amountWithdrawn);
+    }
+
+    /**
+        @notice Withdraw all gOHM from all donor positions
+    */
+    function withdrawAll() external {
+        if (withdrawDisabled) revert YieldDirector_WithdrawalsDisabled();
+
+        uint256[] storage depositIds = depositorIds[msg.sender];
```
Paul4912, line 176: since you are not modifying depositIds it can be memory. I think it might be more gas efficient to use memory here but you will have to test it. It takes more gas to instantiate the array in memory but if you read from it frequently enough it will overall save more gas than storage.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+        if (depositsLength == 0) revert YieldDirector_NoDeposits();
+
+        uint256 principalTotal = 0;
+
+        for (uint256 index = 0; index < depositsLength; ++index) {
+            DepositInfo storage currDeposit = depositInfo[depositIds[index]];
+
+            principalTotal += currDeposit.principalAmount;
+
+            _withdrawAllPrincipal(currDeposit.id);
```
Paul4912, line 188: if depositIds is memory you can just do depositIds[index] here read to memory is cheaper. But will have to test if overall more gas efficient.

contracts/peripheral/YieldDirectorV2.sol
0xLienid: good call

contracts/peripheral/YieldDirectorV2.sol
0xLienid: not really? I could be convinced either way. on the one hand it makes it so that function implementations seem consistent, but also declaring the function means it doesn't actually save any lines of code

contracts/peripheral/YieldDirectorV2.sol
0xLienid: so change it to `if (amount_ >= _toAgnostic(depositInfo[depositId_].principalAmount)` ?

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+        emit AllWithdrawn(msg.sender, _toAgnostic(principalTotal));
+    }
+
+    /**
+        @notice Get deposited gOHM amounts for specific recipient (updated to current index
+                based on sOHM equivalent amount deposit)
+        @param donor_ Address of user donating yield
+        @param recipient_ Address of user receiving donated yield
+    */
+    function depositsTo(address donor_, address recipient_) external view returns (uint256) {
```
Paul4912, line 202: would depositedTo be a better name? Or is frontend already used to depositsTo we don't want to change.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+    function getAllDeposits(address donor_) external view returns (address[] memory, uint256[] memory) {
+        uint256[] storage depositIds = depositorIds[donor_];
+
+        uint256 len = depositIds.length == 0 ? 1 : depositIds.length;
+
+        address[] memory addresses = new address[](len);
+        uint256[] memory agnosticDeposits = new uint256[](len);
+
+        if (depositIds.length == 0) {
+            addresses[0] = address(0);
```
Paul4912, line 246: do we need this? i though address(o) and 0 are default values for address and unsigned integers

Paul4912: i guess its a external view functions so gas isn't really that important here

Paul4912: think you can leave it, might be more clear what it does if you leave it.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,491 @@
...
+        @param depositId_ Deposit ID to withdraw gOHM yield from
+    */
+    function _redeem(uint256 depositId_) internal returns (uint256 amountRedeemed) {
+        if (redeemDisabled) revert YieldDirector_RedeemsDisabled();
+        if (recipientLookup[depositId_] != msg.sender) revert YieldDirector_NotYourDeposit();
+
+        amountRedeemed = _redeemYield(depositId_);
+        if (amountRedeemed == 0) revert YieldDirector_NoRedeemableBalance();
+
+        uint256[] storage receiptIds = recipientIds[msg.sender];
```
Paul4912, line 426: cant this be inside the if statement

test/tyche/YieldDirector.test.ts
```diff
@@ -521,4 +521,8 @@ describe("YieldDirector", async () => {
         await expect(allDeposits[0][0]).is.equal(bob.address);
         await expect(allDeposits[1][0]).is.equal(principal);
     });
+
+    it("test deposit gas", async () => {
```
Paul4912, line 525: accident?

0xLienid:
@Paul4912 No, was to get the gas reading on that deposit to compare with v2's deposit and depositSohm

test/YieldDirectorV2.test.js
```diff
@@ -0,0 +1,817 @@
...
+    it("can't access all deposits per recipient", async () => {
+        const principal = `1${e18}`;
+        await tyche.deposit(principal, bob.address);
+
+        // Since entries in depositInfo are siloed by depositor, funds can't be exploited like this
+        await tyche.connect(alice).deposit(BigNumber.from("1"), bob.address);
+
+        const balanceBefore = await gOhm.balanceOf(alice.address);
+
+        await tyche.connect(alice).withdrawPrincipal("1", principal);
```
Paul4912, line 261: to attempt to steal all the funds the amount you should put in withdrawPrincipal should be 1${e18} right. Thats how the exploit is done to old contract.

test/YieldDirectorV2.test.js
```diff
@@ -0,0 +1,817 @@
...
+        await triggerRebase();
+
+        const allDeposits = await tyche.getAllDeposits(deployer.address);
+        await expect(allDeposits[0].length).is.equal(2);
+        await expect(allDeposits[0][0]).is.equal(bob.address);
+        await expect(allDeposits[1][0]).is.equal(await gOhm.balanceTo(`10${e9}`));
+    });
+
+    it("test deposit gas", async () => {
+        await tyche.deposit(`1${e18}`, bob.address);
```
Paul4912, line 801: should these be left in the pr after done testing? it kinda makes the test suite run longer because the beforeAll is quite long.

test/YieldDirectorV2.test.js
```diff
@@ -0,0 +1,817 @@
...
+        const donationInfo = await tyche.depositInfo("0");
+
+        const withdrawableBalance = await gOhm.balanceTo(donationInfo.principalAmount);
+        await tyche.withdrawPrincipal("0", withdrawableBalance);
+
+        const balanceAfter1 = await gOhm.balanceOf(deployer.address);
+        await expect(balanceAfter1.sub(balanceBefore)).is.equal("999000999200799200"); // this seems to be a bit larger than the usual precision error
+
+        const redeemable = await tyche.redeemableBalance("0");
+        await tyche.connect(bob).redeemYield("0");
```
Paul4912, line 389: do you have a test that test this closes the deposit after withdrawing all? If not heres a decent place to put it.

test/YieldDirectorV2.test.js
```diff
@@ -0,0 +1,817 @@
...
+
+        await expect(await tyche.redeemableBalance("0")).is.equal("0");
+
+        // Second rebase
+        await triggerRebase();
+
+        await expect(await tyche.redeemableBalance("0")).is.equal(await gOhm.balanceTo("10000000"));
+
+        const withdrawableBalance = await gOhm.balanceTo(donationInfo.principalAmount);
+        await tyche.withdrawPrincipal("0", withdrawableBalance);
```
Paul4912, line 599: for one of these tests instead of putting withdrawableBalance here want to try put a super large number to see if it handles it by withdrawing everything?

test/YieldDirectorV2.test.js
Paul4912: oh yea oops

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,488 @@
...
+
+        for (uint256 index = 0; index < receiptIds.length; ++index) {
+            agnosticRedeemable += redeemableBalance(receiptIds[index]);
+        }
+
+        return agnosticRedeemable;
+    }
+
+    function getRecipientIds(address recipient_) external view returns (uint256[] memory) {
+        return recipientIds[recipient_];
```
Paul4912, line 318: this mapping is public why is this needed? do i need this for mine as well?

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,488 @@
...
+    /************************
+     * Recipient Functions
+     ************************/
+
+    /**
+        @notice Get redeemable gOHM balance of a specific deposit
+        @param depositId_ Deposit ID for this donation
+    */
+    function redeemableBalance(uint256 depositId_) public view returns (uint256) {
+        DepositInfo memory currDeposit = depositInfo[depositId_];
```
Paul4912, line 296: have u tested using memory here instead of storage makes it more gas efficient?
i remember mucking around with the old yieldirector with doing stuff like this and it actually made it use more gas. 

From my understanding instantiating a memory costs more gas than instantiating storage since storage is a pointer, memory is an actual copy. But memory becomes more efficient if we perform alot of reads on it. 

So i wouldn't be surprised if totalRedeemableBalance() is more gas efficient using memory but im not so sure about this function since it reads it twice.

Paul4912 (approved):

0xLienid:
@Paul4912 made the getter function because in the public mapping it doesn't return the whole array. You pass two parameters to it, an address and an index, and it only returns that index in the recipient's ID array. So the getter function lets you return the whole array externally.

I did not test it for gas, will test today

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,488 @@
...
+        @notice Deposit additional sOHM, wrap to gOHM, and update deposit record
+        @param depositId_ Deposit ID to direct additional gOHM to
+        @param amount_ Amount of new sOHM debt issued from donor to recipient
+    */
+    function addToSohmDeposit(uint256 depositId_, uint256 amount_) external {
+        uint256 gohmAmount = _toAgnostic(amount_);
+        _increaseDeposit(depositId_, gohmAmount);
+
+        IERC20(sOHM).safeTransferFrom(msg.sender, address(this), amount_);
+        IERC20(sOHM).approve(address(staking), amount_);
```
kushal256, line 131: Use safeApprove?  Also consider doing infinite approval in constructor to save gas

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,488 @@
...
+
+        for (uint256 index = 0; index < depositIds.length; ++index) {
+            uint256 id = depositIds[index];
+
+            if (recipientLookup[id] == recipient_) {
+                return _toAgnostic(depositInfo[id].principalAmount);
+            }
+        }
+
+        return 0;
```
kushal256, line 201: Should this emit an error instead?   Would indicate that the recipient was not found

test/YieldDirectorV2.test.js
```diff
@@ -0,0 +1,809 @@
...
+const { BigNumber } = require("ethers");
+const { expect } = require("chai");
+//const { FakeContract, smock } = require("@defi-wonderland/smock");
+
+const { utils } = require("ethers");
+
+const e9 = "000000000";
+const e18 = "000000000000000000";
+
+describe("YieldDirectorV2", async () => {
```
kushal256, line 11: Loving these tests!

contracts/peripheral/YieldDirectorV2.sol
ind-igo: On second look, I think its ok, to stay consistent with the rest of the internal functions.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,485 @@
...
+        emit Withdrawn(msg.sender, recipientLookup[depositId_], amountWithdrawn);
+    }
+
+    /**
+        @notice Redeem available gOHM yield from a specific deposit
+        @param depositId_ Deposit ID to withdraw gOHM yield from
+    */
+    function _redeem(uint256 depositId_) internal returns (uint256 amountRedeemed) {
+        if (redeemDisabled) revert YieldDirector_RedeemsDisabled();
+        if (recipientLookup[depositId_] != msg.sender) revert YieldDirector_NotYourDeposit();
```
ind-igo, line 416: Should rename this error to `YieldDirector_NotYourYield` since NotYourDeposit is covered in the new YieldSplitter

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,485 @@
...
+
+        amountRedeemed = _redeemYield(depositId_);
+        if (amountRedeemed == 0) revert YieldDirector_NoRedeemableBalance();
+
+        if (depositInfo[depositId_].principalAmount == 0) {
+            _closeDeposit(depositId_);
+
+            uint256[] storage receiptIds = recipientIds[msg.sender];
+            for (uint256 i = 0; i < receiptIds.length; ++i) {
```
ind-igo, line 425: Cache the length since it is used again below.

contracts/peripheral/YieldDirectorV2.sol
```diff
@@ -0,0 +1,485 @@
...
+        if (redeemDisabled) revert YieldDirector_RedeemsDisabled();
+
+        amountRedeemed = 0;
+
+        // Have to read the IDs into memory because with each redemption
+        // an ID is removed from the state array
+        uint256[] memory receiptIds = recipientIds[msg.sender];
+
+        for (uint256 index = 0; index < receiptIds.length; ++index) {
+            uint256 currRedemption = _redeem(receiptIds[index]);
```
ind-igo, line 454: This is extremely expensive because of a loop is used everytime you call this function, potentially O(n^2). I would recreate the redeem logic here instead, by collecting all the redemption information, then deleting everyhting all at once.

ind-igo (changes requested):

contracts/peripheral/YieldDirector.sol
```diff
@@ -3,355 +3,489 @@ pragma solidity ^0.8.10;
...
+
+        amountRedeemed = 0;
+
+        uint256[] storage receiptIds = recipientIds[msg.sender];
+        uint256 idsLength = receiptIds.length;
+
+        // We iterate through the array back to front so that we can delete
+        // elements from the array without changing the locations of any
+        // entries that have not been checked yet
+        for (uint256 index = idsLength; index > 0; index--) {
```
Paul4912, line 471: for better readability why not instead of:
for (uint256 index = idsLength; index > 0; index--)
we do:
for (uint256 index = idsLength-1; index >= 0; index--)

Then inside your loop you don't need to reference index - 1. You can just reference index.

contracts/peripheral/YieldDirector.sol
```diff
@@ -3,355 +3,489 @@ pragma solidity ^0.8.10;
...
-        return (amount_ * index_) / 1e9;
+        @notice Redeem all available gOHM yield from the vault
+    */
+    function _redeemAll() internal returns (uint256 amountRedeemed) {
+        if (redeemDisabled) revert YieldDirector_RedeemsDisabled();
+
+        amountRedeemed = 0;
+
+        uint256[] storage receiptIds = recipientIds[msg.sender];
+        uint256 idsLength = receiptIds.length;
```
Paul4912, line 466: don't think we need to put this variable to memory because since you looping backwards in your for loop this is referenced only once at start and never again. 

Unlike in other for loops where it constantly checks index < length as condition

Paul4912 (approved):

ind-igo (approved):

PR #211, opened 2022-02-13
feat: Common yieldsplitter logic

Paul4912:
Motivation: 
Both pull requests, https://github.com/OlympusDAO/olympus-contracts/pull/208 and https://github.com/OlympusDAO/olympus-contracts/pull/186 use this contract. So i though it would be good idea to agree on the same one so marge it in so we can work on top of it.

Notes for Lienid:
I have decided to remove:
if (userDeposit.principalAmount == 0) {
closeDeposit(id);
}
from _redeemYield()

Reason: I think this should be handled by implementation because I can see some implementations wanting better control of when they close a deposit rather than the last yield withdrawal.

I think this should be the version of YieldSplitter to work off.

### DIFF
--- a/contracts/types/YieldSplitter.sol
+++ b/contracts/types/YieldSplitter.sol
@@ -1,0 +1,141 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity ^0.8.10;
+
+import {IERC20} from "../interfaces/IERC20.sol";
+import {IgOHM} from "../interfaces/IgOHM.sol";
+import {SafeERC20} from "../libraries/SafeERC20.sol";
+
+/**
+    @title YieldSplitter
+    @notice Abstract contract that allows users to create deposits for their gOHM and have
+            their yield claimable by the specified recipient party. This contract's functions
+            are designed to be as generic as possible. This contract's responsibility is
+            the accounting of the yield splitting. All other logic such as error handling,
+            emergency controls, sending and recieving gOHM is up to the implementation of
+            this abstract contract to handle.
+ */
+abstract contract YieldSplitter {
+    using SafeERC20 for IERC20;
+
+    address public immutable gOHM;
+
+    struct DepositInfo {
+        uint256 id;
+        address depositor;
+        uint256 principalAmount; // Total amount of sOhm deposited as principal, 9 decimals.
+        uint256 agnosticAmount; // Total amount deposited priced in gOhm. 18 decimals.
+    }
+
+    uint256 public idCount;
+    mapping(uint256 => DepositInfo) public depositInfo; // depositId -> DepositInfo
+    mapping(address => uint256[]) public depositorIds; // address -> Array of the deposit id's deposited by user
+
+    /**
+        @notice Constructor
+        @param gOHM_ Address of gOHM.
+    */
+    constructor(address gOHM_) {
+        gOHM = gOHM_;
+    }
+
+    /**
+        @notice Create a deposit.
+        @param depositor_ Address of depositor
+        @param amount_ Amount in gOhm. 18 decimals.
+    */
+    function _deposit(
+        address depositor_,
+        uint256 amount_
+    ) internal returns (uint256 depositId) {
+        depositorIds[depositor_].push(idCount);
+
+        depositInfo[idCount] = DepositInfo({
+            id: idCount,
+            depositor: depositor_,
+            principalAmount: IgOHM(gOHM).balanceFrom(amount_),
+            agnosticAmount: amount_
+        });
+
+        depositId = idCount;
+        idCount++;
+    }
+
+    /**
+        @notice Add more gOhm to the depositor's principal deposit.
+        @param id_ Id of the deposit.
+        @param amount_ Amount of gOhm to add. 18 decimals.
+    */
+    function _addToDeposit(uint256 id_, uint256 amount_) internal {
+        DepositInfo storage userDeposit = depositInfo[id_];
+        userDeposit.principalAmount += IgOHM(gOHM).balanceFrom(amount_);
+        userDeposit.agnosticAmount += amount_;
+    }
+
+    /**
+        @notice Withdraw part of the principal amount deposited.
+        @param id_ Id of the deposit.
+        @param amount_ Amount of gOHM to withdraw.
+    */
+    function _withdrawPrincipal(uint256 id_, uint256 amount_) internal {
+        DepositInfo storage userDeposit = depositInfo[id_];
+        userDeposit.principalAmount -= IgOHM(gOHM).balanceFrom(amount_); // Reverts if amount > principal due to underflow
+        userDeposit.agnosticAmount -= amount_;
+    }
+
+    /**
+        @notice Withdraw all of the principal amount deposited.
+        @param id_ Id of the deposit.
+        @return amountWithdrawn : amount of gOHM withdrawn. 18 decimals.
+    */
+    function _withdrawAllPrincipal(uint256 id_) internal returns (uint256 amountWithdrawn) {
+        DepositInfo storage userDeposit = depositInfo[id_];
+        amountWithdrawn = IgOHM(gOHM).balanceTo(userDeposit.principalAmount);
+        userDeposit.principalAmount = 0;
+        userDeposit.agnosticAmount -= amountWithdrawn;
+    }
+
+    /**
+        @notice Redeem excess yield from your deposit in sOHM.
+        @param id_ Id of the deposit.
+        @return amountRedeemed : amount of yield redeemed in gOHM. 18 decimals.
+    */
+    function _redeemYield(uint256 id_) internal returns (uint256 amountRedeemed) {
+        DepositInfo storage userDeposit = depositInfo[id_];
+
+        amountRedeemed = _getOutstandingYield(userDeposit.principalAmount, userDeposit.agnosticAmount);
+        userDeposit.agnosticAmount = IgOHM(gOHM).balanceTo(userDeposit.principalAmount);
+    }
+
+    /**
+        @notice Close a deposit. Remove all information in both the deposit info, depositorIds and recipientIds.
+        @param id_ Id of the deposit.
+        @dev Internally for accounting reasons principal amount is stored in 9 decimal OHM terms. 
+        Since most implementations will work will gOHM, principal here is returned externally in 18 decimal gOHM terms.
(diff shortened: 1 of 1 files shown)


contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,141 @@
...
+        depositId = idCount;
+        idCount++;
+    }
+
+    /**
+        @notice Add more gOhm to the depositor's principal deposit.
+        @param id_ Id of the deposit.
+        @param amount_ Amount of gOhm to add. 18 decimals.
+    */
+    function _addToDeposit(uint256 id_, uint256 amount_) internal {
```
0xLienid, line 68: should we be consistent in spacing of parameters just for consistent readability. _deposit function has them on multiple lines. don't know what's better, not the best with readability

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,141 @@
...
+        userDeposit.principalAmount -= IgOHM(gOHM).balanceFrom(amount_); // Reverts if amount > principal due to underflow
+        userDeposit.agnosticAmount -= amount_;
+    }
+
+    /**
+        @notice Withdraw all of the principal amount deposited.
+        @param id_ Id of the deposit.
+        @return amountWithdrawn : amount of gOHM withdrawn. 18 decimals.
+    */
+    function _withdrawAllPrincipal(uint256 id_) internal returns (uint256 amountWithdrawn) {
```
0xLienid, line 90: this is awesome, will be super helpful

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,141 @@
...
+        userDeposit.principalAmount = 0;
+        userDeposit.agnosticAmount -= amountWithdrawn;
+    }
+
+    /**
+        @notice Redeem excess yield from your deposit in sOHM.
+        @param id_ Id of the deposit.
+        @return amountRedeemed : amount of yield redeemed in gOHM. 18 decimals.
+    */
+    function _redeemYield(uint256 id_) internal returns (uint256 amountRedeemed) {
```
0xLienid, line 102: would there be any use cases of partial redemptions to include in a separate function?

Paul4912: Can't think of any use case but it might be one in future.

From version maintenance POV i think its ok for us to modify this base contract by adding functions. But probably less ok if we modify existing functions. Because if we redeploy an existing implementation with new version and it has an extra internal functions its still fine but deploying with modified base contract function will produce diff results.

So im ok holding off adding it until we need it.

contracts/types/YieldSplitter.sol
Paul4912: We should keep consistent but i dont really care which one I choose. I just ran prettier format and it made it flat again. Idk why it was multiple lines b4. I'l just let the auto format choose whatever it thinks best.

Paul4912: Probably does flat one line unless theres > x number of parameters then it probably expands to multiple lines.

ind-igo:
Looks good. Should we add `_toAgnostic` and `_fromAgnostic` here, so that its not an extra call to gOHM?

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,138 @@
...
+    uint256 public idCount;
+    mapping(uint256 => DepositInfo) public depositInfo; // depositId -> DepositInfo
+    mapping(address => uint256[]) public depositorIds; // address -> Array of the deposit id's deposited by user
+
+    /**
+        @notice Constructor
+        @param gOHM_ Address of gOHM.
+    */
+    constructor(address gOHM_) {
+        gOHM = gOHM_;
```
ind-igo, line 38: I would remove this and just have the _toAgnostic and _fromAgnostic in this file using IsOHM's index. Maybe even don't use IsOHM directly and just define an interface at the top of this file:
```
interface IsOHM {
    function index() external view returns (uint256 index);
}```

File: contracts/types/YieldSplitter.sol [hunk 5]
```diff
@@ -0,0 +1,138 @@
...
+        idCount++;
+    }
+
+    /**
+        @notice Add more gOhm to the depositor's principal deposit.
+        @param id_ Id of the deposit.
+        @param amount_ Amount of gOhm to add. 18 decimals.
+    */
+    function _addToDeposit(uint256 id_, uint256 amount_) internal {
+        DepositInfo storage userDeposit = depositInfo[id_];
```
ind-igo, line 66: Should add errors for checking if deposit ID matches the msg.sender. Its a small lift and can prevent a lot of headaches in the future from inheriting contracts.

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,138 @@
...
+        userDeposit.agnosticAmount += amount_;
+    }
+
+    /**
+        @notice Withdraw part of the principal amount deposited.
+        @param id_ Id of the deposit.
+        @param amount_ Amount of gOHM to withdraw.
+    */
+    function _withdrawPrincipal(uint256 id_, uint256 amount_) internal {
+        DepositInfo storage userDeposit = depositInfo[id_];
```
ind-igo, line 77: Again, add errors here. The withdrawal functions should be protected as much as possible. Too much of a risk for draining funds.

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,167 @@
...
+        userDeposit.agnosticAmount -= amount_;
+    }
+
+    /**
+        @notice Withdraw all of the principal amount deposited.
+        @param id_ Id of the deposit.
+        @return amountWithdrawn : amount of gOHM withdrawn. 18 decimals.
+    */
+    function _withdrawAllPrincipal(uint256 id_) internal returns (uint256 amountWithdrawn) {
+        if (depositInfo[id_].depositor != msg.sender) revert YieldSplitter_NotYourDeposit();
```
0xLienid, line 97: would this not read the YieldDirectorV2 address as msg.sender? or is it different when inheriting abstracts?

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,167 @@
...
+        userDeposit.agnosticAmount += amount_;
+    }
+
+    /**
+        @notice Withdraw part of the principal amount deposited.
+        @param id_ Id of the deposit.
+        @param amount_ Amount of gOHM to withdraw.
+    */
+    function _withdrawPrincipal(uint256 id_, uint256 amount_) internal {
+        if (depositInfo[id_].depositor != msg.sender) revert YieldSplitter_NotYourDeposit();
```
0xLienid, line 84: would this not read the YieldDirectorV2 address as msg.sender? or is it different when inheriting abstracts?

contracts/types/YieldSplitter.sol
Paul4912: that only happens when another contract calls this contract. When you inherit the base contract you just combine and become as one sorta. So this will be the same msg.sender as YieldDirectorV2's msg.sender.

0xLienid: very good to know

Paul4912: The only concern I have with this check is if we want to do something like allow another contract to withdraw for the user. I see this happening being a thing more for redeeming yield tho so I don't see this happening with withdraw.

Do you think we will have implementations that would want to allow another contract to withdraw on behalf of a user? if so we should probably remove this and make implementation handle this check.

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,167 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity ^0.8.10;
+
+import {IERC20} from "../interfaces/IERC20.sol";
+import {IgOHM} from "../interfaces/IgOHM.sol";
+import {SafeERC20} from "../libraries/SafeERC20.sol";
+
+interface IOHMIndexWrapper {
```
ind-igo, line 8: Would add an `@dev` comment above this explaining why this is needed vs IsOHM or something. Maybe:

```
This interface is so that we can wrap cross-chain oracles to also feed an index without needing to adhere completely to IsOHM.
```

contracts/types/YieldSplitter.sol
ind-igo: I think this contract should handle that case, since it is a base contract to be inherited. I see this as a building block for other systems to be built on, not just the two of them that we have currently.

ind-igo (commented):
almost done. just superficial changes and i think its ready (assuming tests work out)

Paul4912:
> almost done. just superficial changes and i think its ready (assuming tests work out)

i have tests on other pr but they need to be updated. want to get this in so me first so its easier for lienid and i to work off same thing. Will fix up tests after i rebase on this merged in.

contracts/types/YieldSplitter.sol
0xLienid: so for example in the Odyssey buy-now-pay-later idea, we would build that directly off of this rather than building something that interacts through say YieldDirector, and thus this check still works?

Paul4912: Yea this check will work as long as the depositor will be calling the same contract they deposited to. 

So in the case of an odyssey contract that will build a contract that implements this base contract. As long as depositor is interacting with the contract that implements this contract will be all good. But if theres another contract that will interact with the odyssey contract wants to withdraw on behalf of user would stuff up.

contracts/types/YieldSplitter.sol
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /**
+        @notice Convert agnostic gOHM value at current index to flat sOHM value
+        @dev Agnostic value earns rebases. sOHM amount is gOHMamount * rebase_index.
+             1e18 is because sOHM has 9 decimals, gOHM has 18 and index has 9.
+     */
+    function _fromAgnostic(uint256 amount_) internal view returns (uint256) {
+        return (amount_ * (indexWrapper.index())) / 1e18;
+    }
```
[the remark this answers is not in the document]
Paul4912: I think we can add that when we need it. Right now your yieldDirectory and my yieldsteamer doesn't need it.

contracts/types/YieldSplitter.sol
ind-igo: typo here, sorry. my fault 😅

`to also feed an index` should actually be `to feed an index` (the also shouldnt be there)

PR #214, opened 2022-02-17
build: enable codeql security scan for js files

ivelin:


Paul4912:
Is this needed for smart contract repo since the only js files would be deploy scripts and tests?

ivelin:
Good question. @gordob suggested that we add codeql to this repo as it won't hurt to scan js files.
Solidity code scanning is not available for CodeQL yet, but if and when it is, we can turn it on.

On a related note, @gordob found this automated smart contract scanning tool. One of their reps was promoting it in the DAO discord server. Looks legit. Built by the OpenZeppelin team. a16z funded.
https://explorer.forta.network/

PR #247, opened 2022-03-05
ci: deployment script for `TreasuryExtender`

mirru2532:
Deployment script for `TreasuryExtender`.


mirru2532:
Closing this PR as we the `hardhat-deploy` system needs to be checked out by properly configuring json's so we don't accidentally  redeploy a contract as I have redeployed today.

PR #252, opened 2022-03-08
Propose interfaces for incur debt

tonypiano:
IIncurDebt is the full-blown solution that manages the LP position.
IIncurDebtTokemak is a simpler version that just gives the borrower OHM.

### DIFF
--- a/contracts/interfaces/IIncurDebtTokemak.sol
+++ b/contracts/interfaces/IIncurDebtTokemak.sol
@@ -77,7 +77,7 @@ interface IIncurDebtTokemak {
     function setBorrowerDebtLimit(address _account, uint256 _debtLimit) external;
 
     /**
-     * @notice removes LP position, repays debt and returns tokens to owner
+     * @notice repays debt using collateral and returns remaining tokens to owner
      * - onlyOwner (or governance)
      * @param _account the address that will interact with contract
      * @param _to where to send remaining gOHM
@@ -85,5 +85,13 @@ interface IIncurDebtTokemak {
     function forceRepay(
         address _account,
         address _to
-    ) external returns (uint256);
+    ) external;
+
+    /**
+     * @notice seize and burn _accounts collateral and forgive debt
+     * - will burn all collateral, including excess of debt
+     * - onlyGovernance
+     * @param _account the account to seize
+     */
+    function seize(address _account) external;
 }
--- a/contracts/interfaces/IIncurDebtTokemak.sol
+++ b/contracts/interfaces/IIncurDebtTokemak.sol
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity >=0.8.4;
+
+/// @notice simplified IncurDebt to work as a Tokemak controller
+/// @dev this contract must be given OHMDEBTOR on the treasury
+interface IIncurDebtTokemak {
+    /* ========== TOKEMAK FUNCTIONS ========== */
+
+    /**
+     * @notice deposits gOHM to use as collateral
+     * - msg.sender must be whitelisted
+     * - this contract must have been approved _amount of gOHM
+     * @dev will unwrap and hold as sOHM in this contract
+     * @param _gOHMAmount amount of gOHM
+     */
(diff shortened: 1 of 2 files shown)


contracts/interfaces/IIncurDebtTokemak.sol
```diff
@@ -0,0 +1,97 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity >=0.8.4;
+
+/// @notice simplified IncurDebt to work as a Tokemak controller
+/// @dev this contract must be given OHMDEBTOR on the treasury
+interface IIncurDebtTokemak {
```
0xJFry4, line 6: Since we're going with MVP this interface will be for more than Tokemak, right?

contracts/interfaces/IIncurDebtTokemak.sol
```diff
@@ -0,0 +1,97 @@
...
+     * @param _ohmAmount amount of OHM to borrow
+     */
+    function repay(uint256 _ohmAmount) external;
+
+    /**
+     * @notice gets available OHM to borrow for account
+     * @param _account account to inspect
+     * @return amount OHM available to borrow
+     */
+    function getAvailableToBorrow(address _account) external returns(uint256);
```
0xJFry4, line 40: Should be a view?

contracts/interfaces/IIncurDebtTokemak.sol
```diff
@@ -0,0 +1,97 @@
...
+     */
+    function setBorrowerDebtLimit(address _account, uint256 _debtLimit) external;
+
+    /**
+     * @notice repays debt using collateral and returns remaining tokens to owner
+     * - onlyOwner (or governance)
+     * @param _account the address that will interact with contract
+     * @param _to where to send remaining gOHM
+     */
+    function forceRepay(
```
0xJFry4, line 85: Would it be good to have a method that combines forceRepay and revokeBorrower? In the event that Governance decides to force quit a partner without burning their collateral it would require 2 separate contract calls. (mostly thinking out loud)

contracts/interfaces/IIncurDebtTokemak.sol
```diff
@@ -0,0 +1,97 @@
...
+    function repay(uint256 _ohmAmount) external;
+
+    /**
+     * @notice gets available OHM to borrow for account
+     * @param _account account to inspect
+     * @return amount OHM available to borrow
+     */
+    function getAvailableToBorrow(address _account) external returns(uint256);
+
+    /* ========== MANAGEMENT FUNCTIONS ========== */
```
0xJFry4, line 42: We also need a way for Policy to set LTV. Currently we've discussed this being 100%, but I think it would be good to have flexibility in the future. This would meant hat borrows need to check LTV against how much the partner wants to borrow.

contracts/interfaces/IIncurDebtTokemak.sol
```diff
@@ -0,0 +1,97 @@
...
+     * - this contract must have been approved _amount of gOHM
+     * @dev will unwrap and hold as sOHM in this contract
+     * @param _gOHMAmount amount of gOHM
+     */
+    function deposit(uint256 _gOHMAmount) external;
+
+    /**
+     * @notice deposits gOHM to use as collateral
+     * - msg.sender must be whitelisted
+     * - _ohmAmount must be <= availableDebt
```
0xJFry4, line 21: Also needs to check LTV %

0xJFry4 (changes requested):
Looks great. Some minor comments

Paul4912:
So i caught up with the messages on the discord. Why are we making a version where we manage the LP's and one where we just give borrower OHM? Are both going to be used or are we tossing up options?

Paul4912:
I think it might be better to combine the interfaces and have 1 contract do both roles. 
Benefit:
-We don't need to keep track of our total debt limits from two different contracts. 
-I anticipate minimal friction since we just need to add another role for permissioning for tokemak and FiatDAO to use some restricted functions
-One contract to deploy keep track of

Cons:
-unforeseen smart contract friction

PARABRAHMAN0:
Contact me if you want this reopened.

PR #255, opened 2022-03-12
Onsen allocators first draft

calusaca:
As I understand Onsen works like this:

Once we deposit SLP tokens into a specific pool (_update(uint256 id)), we will need to manage the sushi tokens (rewards) which we can be staked, if we do stake we get back xSushi which represent the sushi staked.

So based on this I have the following questions:

1- How should we handle the sushi rewards, should we lock those sushi tokens in the "sushi bar"? 
2- What would be the strategy to calculate gains from the sushi rewards if there is any?
2a - Should we swap it to the current LP token and return that as the gains?

### DIFF
--- a/contracts/allocators/OnsenAllocatorV2.sol
+++ b/contracts/allocators/OnsenAllocatorV2.sol
@@ -1,0 +1,139 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity ^0.8.10;
+import "../libraries/Address.sol";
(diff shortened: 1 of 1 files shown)


Paul4912:
this same as https://github.com/OlympusDAO/olympus-contracts/pull/265 right?
if so you can close this pr now

calusaca:
> this same as #265 right? if so you can close this pr now

Yes this was a previous one, I just closed this one, thanks

PR #256, opened 2022-03-12
doc: add links to doc files and add allocator addresses

PARABRAHMAN0:
Docs have been added in second commit and addresses in first commit.


PARABRAHMAN0:
The above PAT is no longer in use.

ind-igo (approved):

PR #259, opened 2022-03-13
Lobi allocator, tests, and deploy script

0xLienid:


test/allocators/LobiAllocator.test.ts
```diff
@@ -0,0 +1,452 @@
...
+                amount
+            );
+
+            await expect(() => allocator.update(1)).to.changeTokenBalance(
+                lobi,
+                allocator,
+                BigNumber.from(0).sub(amount)
+            );
+
+            const mAllocator: LobiAllocator = await factory.deploy(
```
ind-igo, line 412: if this deploy is being done a lot then just make this into a function you call.

ind-igo (changes requested):

PR #264, opened 2022-03-17
modify yieldsplitter to allow authorized contracts to redeem onbehalf of recipients

Paul4912:
Motivation: To future proof our contracts we might need external contracts to redeem deposits on-behalf of recipients. An example would be an aggregator contract to redeem a recipient's yield from multiple yield-splitter implementations.

Changes: Add giving and revoking permission to base yieldsplitter contract.
Add virtual functions for implementation to redeem on behalf of users and mandatory view functions
Moved Olympus authority to base contract.
Added some new tests.


PARABRAHMAN0:
Since contract is in audit this change can't be applied any longer.

PR #265, opened 2022-03-17
Onsen allocator

calusaca:
Questions:

1.  About update, once we stake the sushi tokens (rewards), we get back xSushi tokens, what should we do with those tokens? Send those back to the treasury?

2.  For deallocate, the index will represent the id of the tokens to be withdraw?

Asuming the index of the amounts [] can be use as parameter for the mapping tokenIds
Example:
amounts => [20,50,150]
So the values that will be used to retrieve the id used in onsen will be 0, 1 , 2

3. About deallocate, should we send tokens to treasury once those are un-stake?

4. About deallocate, should we send reward tokens to treasury once those are un-stake?

5. For deactivate haven’t found a way to complete this without the TokenId used in Onsen that is stored on mapping tokenIds, any ideas?


contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+        return "OnsenAllocator";
+    }
+
+    /* ========== INTERNAL FUNCTIONS ========== */
+
+    /**
+     * @notice stake sushi rewards if enter is true else return funds to treasury.
+     * @param _stake bool
+     */
+    function enterSushiBar(bool _stake) internal {
```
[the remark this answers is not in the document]
ind-igo: Also, why isthis internal? The only time you call this is in update and pass in true. So you would never be able to unstake?

Also, the naming does not really work. if  `_stake` is false, you aren't entering the sushi bar, you're exiting.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+    uint256 rewardDebt;
+}
+
+interface IMasterChef {
+    function pendingSushi(uint256 _pid, address _user) external view returns (uint256);
+
+    function deposit(uint256 _pid, uint256 _amount) external;
+
+    function withdraw(uint256 _pid, uint256 _amount) external;
+
```
Paul4912, line 19: Theres a subfolder for interfaces in the allocators folder. Think this belongs there.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+    /* ========== INTERNAL FUNCTIONS ========== */
+
+    /**
+     * @notice stake sushi rewards if enter is true else return funds to treasury.
+     * @param _stake bool
+     */
+    function enterSushiBar(bool _stake) internal {
+        uint256 balance = IERC20(sushi).balanceOf(address(this));
+        if (balance > 0) {
+            if (!_stake) {
```
ind-igo, line 155: Would check for true case first. It's odd to check for false case here, considering false is the less common case.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+        Asuming the index of the amounts [] can be use as parameter for the mapping tokenIds
+        Example:
+        amounts => [20,50,150]
+        So the values that will be used to retrieve the id used in onsen will be 0, 1 , 2
+    */
+    function deallocate(uint256[] memory amounts) public override onlyGuardian {
+        for (uint256 i; i < amounts.length; i++) {
+            uint256 amount = amounts[i];
+            if (amount > 0) {
+                IMasterChef(masterChef).withdraw(tokenIds[i], amount); // withdraw from Onsen
```
ind-igo, line 115: Is this how you exit or are you calling `enterSushiBar(false)` ?

ind-igo (changes requested):

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+
+        // interactions
+        //Deposit LP token into onsen, if deposit succesfull this address should have in return sushi tokens
+        if (balance > 0) {
+            IERC20(LPtoken).approve(masterChef, balance);
+            IMasterChef(masterChef).deposit(id, balance); // deposit into Onsen
+        }
+
+        // Stake the sushi tokens
+        if (IERC20(sushi).balanceOf(address(this)) > 0) {
```
ind-igo, line 88: No point of checking this, since the same check happens inside `enterSushiBar`

contracts/allocators/OnsenAllocatorV2.sol
calusaca: Interfaces moved to the interface subfolder and the contract was updated to reference the interfaces.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+
+    /* ========== STATE VARIABLES ========== */
+
+    address immutable sushi; // $SUSHI token
+    address immutable xSushi; // $xSUSHI token
+
+    address immutable masterChef; // Onsen contract
+
+    // address immutable treasury;
+    address public constant treasury = 0x9A315BdF513367C0377FB36545857d12e85813Ef; // Olympus Treasury
```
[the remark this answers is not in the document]
calusaca: treasury variable is immutable.
treasury address pass in the constructor.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        //Calculate gains/loss
+        // Retrieve current balance for pool and address
+        UserInfo memory currentUserInfo = IMasterChef(masterChef).userInfo(id, address(this));
+
+        uint256 last = extender.getAllocatorAllocated(id) + extender.getAllocatorPerformance(id).gain;
+
+        if (currentUserInfo.amount >= last) {
+            gain = uint128(currentUserInfo.amount - last);
```
Paul4912, line 81: so since theres no compounding and LP's gain and lose value from being worth more or less due to fees and IL. Realistically we won't ever see any gains or losses right since the amount of LP tokens will always stay the same.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+        AllocatorInitData memory data
+    ) BaseAllocator(data) {
+        require(_chef != address(0));
+        masterChef = _chef;
+        require(_sushi != address(0));
+        sushi = _sushi;
+        require(_xSushi != address(0));
+        xSushi = _xSushi;
+
+        // approve for safety, yes toke is being instantly sent to treasury and that is fine
```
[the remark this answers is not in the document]
calusaca: Comments updated to reflect Sushi and xSushi

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,156 @@
...
+            uint256 amount = amounts[i];
+            if (amount > 0) {
+                IMasterChef(masterChef).withdraw(tokenIds[i], amount); // withdraw from Onsen
+            }
+        }
+    }
+
+    function _deactivate(bool panic) internal override {}
+
+    function _prepareMigration() internal override {}
```
Paul4912, line 104: shouldn't this deallocate everything?

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,156 @@
...
+        // reads
+        uint256 index = tokenIds[id];
+        IERC20 LPtoken = _tokens[index];
+        uint256 balance = LPtoken.balanceOf(address(this));
+
+        // interactions
+        //Deposit LP token into onsen, if deposit succesfull this address should have in return sushi tokens
+        if (balance > 0) {
+            IERC20(LPtoken).approve(masterChef, balance);
+            IMasterChef(masterChef).deposit(id, balance); // deposit into Onsen
```
Paul4912, line 66: the masterchef id is going to be a different number to ids of the deposits in the `TreasuryExtender` right?

contracts/allocators/OnsenAllocatorV2.sol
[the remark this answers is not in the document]
calusaca: > rename to `_enterSushiBar` to denote an internal function
A/ I used this name because it reflects the contract and the specific function that is being called from ONSEN/SUSHI
SushiBar.sol/function enter 
But I can change it to:
function manageSushi(bool ) internal {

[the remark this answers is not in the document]
calusaca: > Also, why isthis internal? The only time you call this is in update and pass in true. So you would never be able to unstake?
> 
> Also, the naming does not really work. if `_stake` is false, you aren't entering the sushi bar, you're exiting.

This is internal because it is handling what to do with sushi token (rewards), once a LP is deposited into ONSEN sushi tokens are transferred to the sender.
The logic behind the _stake was that you could send the sushi tokens back to the treasury or stake them, so the idea of this function was not to un-stake, I will change the function to avoid passing a boolean and always just stake the sushi tokens.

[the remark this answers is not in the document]
calusaca: > rename to `_enterSushiBar` to denote an internal function

I used this name because it reflects the contract and the specific function that is being called from ONSEN/SUSHI
SushiBar.sol/function enter
But I can change it to something like:
function stakeSushi() internal {

Since this function is only going to be doing that.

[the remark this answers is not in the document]
calusaca: Changes applied:
function enterSushiBar(bool _stake) internal {
to
function stakeSushi() internal {

only function is to stake sushi tokens and send to treasury xSushi tokens

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+                IERC20(xSushi).safeTransfer(treasury, xBalance); // transfer xSushi to treasury
+            }
+        }
+    }
+
+    /**
+     *  @notice pending $SUSHI rewards
+     *  @return uint
+     */
+    function pendingSushi(uint256 pid) internal view returns (uint256) {
```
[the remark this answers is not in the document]
calusaca: Function name updated to _pendingSushi

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,174 @@
...
+        xSushi = _xSushi;
+
+        // approve for safety, yes toke is being instantly sent to treasury and that is fine
+        // but to be absolutely safe this one approve won't hurt
+        IERC20(sushi).approve(address(extender), type(uint256).max);
+        IERC20(xSushi).approve(address(extender), type(uint256).max);
+    }
+
+    /**
+     1- Deposit LP token into onsen, if deposit is succesfull we should get in return sushi tokens
```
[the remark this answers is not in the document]
calusaca: Comments updated to follow NatSpec format.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: This function was updated to just stake any sushi tokens available, so check is no longer needed.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: IMasterChef(masterChef).withdraw that's the function to unstake the LP tokens, 
 I am adding the functionality to unstake sushi tokens.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: Code updated to avoid the double-check of sushi balance

contracts/allocators/OnsenAllocatorV2.sol
calusaca: @Paul4912  Yes that's correct there shouldn't be any difference for the LP amount, but not sure if we should still do the comparison? or should we remove it and always return 0 for gain and loss?

contracts/allocators/OnsenAllocatorV2.sol
calusaca: @Paul4912 Yes it does, but I am having an issue trying to get the list of tokenIds to be deallocated, since 
  // From deposit id to the token's id
    mapping(uint256 => uint256) public tokenIds;
    
 is a mapping structure so I cannot iterate over it, and I haven't been able how to figure out which are the active ids from that list that are needed to call the withdraw function. 
 @Paul4912 / @ind-igo / @mirru2532  If you have any ideas please let me know, I will continue trying to figure out a way but so far I don't know.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: @Paul4912 sorry but not following, I am using the id that is passed to the deposit function as the id for the deposit into masterchef
is that the id you are referring to?

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,149 @@
...
+     * 4- Calculate gains/loss
+     */
+    function _update(uint256 id) internal override returns (uint128 gain, uint128 loss) {
+        uint256 index = tokenIds[id];
+        IERC20 LPtoken = _tokens[index];
+        uint256 balance = LPtoken.balanceOf(address(this));
+
+        /// Deposit LP token into onsen, if deposit succesfull this address should have in return sushi tokens
+        if (balance > 0) {
+            IERC20(LPtoken).approve(masterChef, balance);
```
ind-igo, line 66: This does not need to be cast to IERC20 again. LPtoken is already of the type IERC20

calusaca: Code updated to avoid casting to IERC20 again when it is already IERC20

contracts/allocators/OnsenAllocatorV2.sol
PARABRAHMAN0: You need to know if you the LP tokens are constant balance and increasing value or vice versa, and if the former note down the increasing balance, since the LP tokens are being withdrawn from the treasury and this needs to be registered sir. Still, always note down the LP balance please, since there might be a contract failure and you need to report that so the allocator can panic and potentially do something in case where all tokens have not been lost.

contracts/allocators/OnsenAllocatorV2.sol
PARABRAHMAN0: Again here the proper ids need to be used since you cannot repurpose tokenIds.

contracts/allocators/OnsenAllocatorV2.sol
PARABRAHMAN0: You need to hold an iterable array instead of a mapping, tokenIds is as said not for this usecase. Only transforming as mentioned above.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,149 @@
...
+     * @notice stake sushi rewards
+     */
+    function _stakeSushi() internal {
+        uint256 balance = IERC20(sushi).balanceOf(address(this));
+        if (balance > 0) {
+            IERC20(sushi).approve(xSushi, balance);
+            ISushiBar(xSushi).enter(balance); // stake sushi
+
+            uint256 xBalance = IERC20(xSushi).balanceOf(address(this));
+            IERC20(xSushi).safeTransfer(treasury, xBalance); // transfer xSushi to treasury
```
PARABRAHMAN0, line 138: Do we want to explicitly send xSushi?

PARABRAHMAN0 (changes requested):
Done with review.

contracts/allocators/OnsenAllocatorV2.sol
Paul4912: So for the masterchef contract each LP token farm is mapped to an ID. To do anything like deposit you need to input the correct ID that corresponds to the LP token. 

The id in _update(id) is the id corresponding to the deposit id that the allocator and treasury extender references. This is not the same id as the id that we need to put into the masterchef contract made by sushi.

As PARABRAHMAN0 mentioned in another comment you should probably have an array of integers the same length as IERC20[] internal _tokens to store the masterchef id's of all your LP tokens. This way when you construct the _tokens in the constructor you can also construct another array that maps 1 to 1 to easily access their masterchef id's. You won't need a new mapping since it will be sharing the mapping(uint256 => uint256) public tokenIds mapping.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: This function was changed for stakeSushi.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: Gain/Loss balance is still been returned based on the balance for the LP tokens in the Onsen pool.

LP tokens balance is constant, since rewards are awarded with sushi tokens

contracts/allocators/OnsenAllocatorV2.sol
calusaca: Finally I understood how to get the id of the Onsen Pool based on the LP token, I am keeping locally a mapping to easily get the Onsen pool id based on the LP token.
This logic was updated.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,149 @@
...
+     */
+    function _update(uint256 id) internal override returns (uint128 gain, uint128 loss) {
+        uint256 index = tokenIds[id];
+        IERC20 LPtoken = _tokens[index];
+        uint256 balance = LPtoken.balanceOf(address(this));
+
+        /// Deposit LP token into onsen, if deposit succesfull this address should have in return sushi tokens
+        if (balance > 0) {
+            IERC20(LPtoken).approve(masterChef, balance);
+            IMasterChef(masterChef).deposit(id, balance);
```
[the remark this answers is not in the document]
calusaca: Logic updated to search for the Onsen Pool based on the LP token.

PR #273, opened 2022-03-24
Add Slither analysis to pull requests

0xJem:
Slither will be run on each pull request and listed under the "Checks" section.

Screenshot 2022-03-24 at 3 09 59 PM

e.g. https://github.com/0xJem/olympus-contracts/pull/1/checks?check_run_id=5675243069

### DIFF
--- a/.github/workflows/slither.yml
+++ b/.github/workflows/slither.yml
@@ -1,19 +1,18 @@
 name: Slither Analysis
 on: [pull_request]
 jobs:
-  analyze:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - run: pwd
-      - run: ls -al
-      - uses: crytic/slither-action@v0.1.0
-        name: Perform Slither Analysis
-        id: slither
-        continue-on-error: true
-        with:
-         sarif: results.sarif
-      - name: Upload SARIF file
-        uses: github/codeql-action/upload-sarif@v1
-        with:
-          sarif_file: ${{ steps.slither.outputs.sarif }}
+    analyze:
+        runs-on: ubuntu-latest
+        steps:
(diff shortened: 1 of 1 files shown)


PARABRAHMAN0:
Seriously cool!

that is a lot of results though >.> i removed a couple that were obv from the severe ones  / that won't get replaced

I'd check out the name duplications and if not just ignore all and pray

0xJem:
From @ind-igo :
> i think we may not want it as a check on the repo at the moment though. i think moving forward, we want the action setup on new repos

Right now, the slither output is not showing up under the security tab. I think we need to enable code scanning for that to work: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/setting-up-code-scanning-for-a-repository

@ind-igo @appleseed-iii if you have admin access to the repo, can you enable that?

Secondly, I can't see a way to disable a failing check (code scanning results) causing the failure of the entire build. If we can get the security reports running, I can see if it's possible to remove the Slither output as a "check" on the PR.

philipjonsen (approved):

0xJem:
@PARABRAHMAN0 @ind-igo I'm looking at how we can reduce the number of warnings, a concern that was raised earlier. 

One suggestion was to not analyse contracts that have already been deployed. This would require maintaining a list of filenames as an exclude list, and would require devs to remember to update it after deployment. Is that feasible?

The other options is slither's triage mode: https://github.com/crytic/slither/wiki/Usage#triage-mode
This would require a dev to run the command (`slither . --triage-mode`) and tell slither to ignore any warnings that are no longer relevant (or which can't be acted upon). This will get written to a JSON file and committed to the repo, and will prevent the slither action in github from reporting on that issue.

Which would you prefer?

PR #278, opened 2022-03-28
Onsen Allocator update

calusaca:
Added Onsen allocator based on the latest allocators branch.

### DIFF
--- a/contracts/allocators/OnsenAllocatorV2.sol
+++ b/contracts/allocators/OnsenAllocatorV2.sol
@@ -1,0 +1,268 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity ^0.8.10;
+import "../libraries/Address.sol";
+
+///  ========== TYPES ==========
+import "../types/BaseAllocator.sol";
+
+///  ========== INTERFACES ==========
+import "./interfaces/IMasterChef.sol";
+import "./interfaces/ISushiBar.sol";
+
+struct PoolData {
+    bool poolActive;
+    uint256 id;
+}
+
+/**
+ *  @notice Contract deploys liquidity from treasury into the Onsen program,
+ *  earning $SUSHI that can be staked and/or deposited into the treasury.
+ */
+contract OnsenAllocatorV2 is BaseAllocator {
+    using SafeERC20 for IERC20;
+
+    /// ========== STATE VARIABLES ==========
+
+    address public sushi; /// $SUSHI token
+    address public xSushi; /// $xSUSHI token
+    address public masterChef; /// Onsen contract
+
+    address public treasury; /// Olympus Treasury
+
+    mapping(address => PoolData) internal _lpToOnsenId;
+
+    /// ========== CONSTRUCTOR ==========
+
+    constructor(
+        address _chef,
+        address _sushi,
+        address _xSushi,
+        address _treasury,
+        AllocatorInitData memory data
+    ) BaseAllocator(data) {
+        require(_chef != address(0));
+        masterChef = _chef;
+        require(_sushi != address(0));
+        sushi = _sushi;
+        require(_xSushi != address(0));
+        xSushi = _xSushi;
+        require(_treasury != address(0));
+        treasury = _treasury;
+
+        /**  @dev approve for safety, sushi is being staked and xsushi is being instantly sent to treasury and that is fine
+         *   but to be absolutely safe this one approve won't hurt
+         */
+        IERC20(sushi).approve(address(extender), type(uint256).max);
+        IERC20(xSushi).approve(address(extender), type(uint256).max);
+    }
+
+    /**
+     * @notice Find out which LP tokens to be deposited into onsen, then deposit it into onsen pool and stake the sushi tokens that are returned.
+     * 1- Based on the LP token to be deposited, find out the Onsen Pool Id
+     * 2- Deposit LP token into onsen, if deposit is succesfull we should get in return sushi tokens
+     * 2- Stake the Sushi rewards tokens returned from deposit LP tokens and as rewards form onsen LP stake, we should get in return xSushi tokens
+     * 3- Keep the xSushi on the allocator contract for easily deallocation
+     * 4- Calculate gains/loss based on the LP balance on the Onsen Pool.
+     */
+    function _update(uint256 id) internal override returns (uint128 gain, uint128 loss) {
+        uint256 index = tokenIds[id];
+        IERC20 LPtoken = _tokens[index];
+        uint256 balance = LPtoken.balanceOf(address(this));
+
+        /// Deposit LP token into onsen, if deposit succesfull this address should have in return sushi tokens
+        if (balance > 0) {
+            /// Find out if there is a Onsen Pool for the LPToken
+            (bool onsenLPFound, uint256 onsenPoolId) = _findPoolByLP(address(LPtoken));
+
+            if (onsenLPFound) {
+                /// Approve and deposit balance into onsen Pool
+                LPtoken.approve(masterChef, balance);
+                IMasterChef(masterChef).deposit(onsenPoolId, balance, address(this));
+
+                /// Stake the sushi tokens
+                _stakeSushi();
+
+                ///Calculate gains/loss
+                /// Retrieve current balance for pool and address
+                UserInfo memory currentUserInfo = IMasterChef(masterChef).userInfo(onsenPoolId, address(this));
+
+                uint256 last = extender.getAllocatorAllocated(id) + extender.getAllocatorPerformance(id).gain;
+
+                if (currentUserInfo.amount >= last) {
+                    gain = uint128(currentUserInfo.amount - last);
+                } else {
+                    loss = uint128(last - currentUserInfo.amount);
+                }
+            } else {
+                /// If no Onsen pool was found for the LP token then return LP token to the treasury.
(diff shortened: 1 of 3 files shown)


contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,268 @@
...
+        if (balance > 0) {
+            ISushiBar(xSushi).leave(balance); // unstake $xSUSHI
+        }
+    }
+
+    /**
+     *  @notice pending $SUSHI rewards
+     *  @return uint
+     */
+    function _pendingSushi(uint256 pid) internal view returns (uint256) {
```
Paul4912, line 218: this is unused?

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,268 @@
...
+        PoolData memory pd = _lpToOnsenId[LPToken];
+
+        /// If the pool is active return the id else search for the pool id
+        if (pd.poolActive) {
+            return (pd.poolActive, pd.id);
+        } else {
+            /// Call the poolLength function from sushi masterchef v2 and compare against the LP token passed as parameter
+            uint256 poolsLength = IMasterChef(masterChef).poolLength();
+
+            for (uint256 i; i < poolsLength; i++) {
```
Paul4912, line 238: Poollength of masterchef is 355 i think unless im looking at wrong contract. i think this can be done offchain. We can figure out the id mappings offchain using view functions on masterchef and insert them in with a setter function. This would work but im worried it will use alot more gas than we need to.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,268 @@
...
+                /// If found, return the the pool Id from Onsen for the current LPToken.
+                if (LPToken == (address)(IMasterChef(masterChef).lpToken(i))) {
+                    pd = _lpToOnsenId[LPToken];
+                    pd.poolActive = true;
```
Paul4912, line 242: if were changing variables here this not a view function anymore right though the compiler would say something.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,268 @@
...
+     * 3- Keep the xSushi on the allocator contract for easily deallocation
+     * 4- Calculate gains/loss based on the LP balance on the Onsen Pool.
+     */
+    function _update(uint256 id) internal override returns (uint128 gain, uint128 loss) {
+        uint256 index = tokenIds[id];
+        IERC20 LPtoken = _tokens[index];
+        uint256 balance = LPtoken.balanceOf(address(this));
+
+        /// Deposit LP token into onsen, if deposit succesfull this address should have in return sushi tokens
+        if (balance > 0) {
```
Paul4912, line 73: so i read the masterchef contract and it claims your pending sushi when u deposit or withdraw correct?

Not sure what the strategy is financially but i would imagine we have a case where we would wana not deposit or withdraw anything but compound our stack by claiming and staking sushi. 

If we want to do that potentially maybe we can remove the balance > 0 check.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,268 @@
...
+        for (uint256 i; i < _tokens.length; i++) {
+            (bool onsenLPFound, uint256 onsenPoolId) = _findPoolByLP(address(_tokens[i]));
+            if (onsenLPFound) {
+                /// Retrieve current balance for pool and address
+                UserInfo memory currentUserInfo = IMasterChef(masterChef).userInfo(onsenPoolId, address(this));
+                amounts[i] = currentUserInfo.amount;
+            }
+        }
+
+        deallocate(amounts);
```
Paul4912, line 266: This seems unnecessary layer of function calling to me. We can just call withdraw and harvest in our first for loop instead of delegating to this function and looping again.

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,268 @@
...
+        if (balance > 0) {
+            IERC20(sushi).approve(xSushi, balance);
+            ISushiBar(xSushi).enter(balance); // stake sushi
+        }
+    }
+
+    /**
+     * @notice unStake sushi rewards
+     */
+    function _unStakeSushi() internal {
```
Paul4912, line 207: fairly unimportant but if u using camel case the first s is lowercase because unstake is a whole word.

contracts/allocators/OnsenAllocatorV2.sol
PARABRAHMAN0: +1

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,268 @@
...
+                    /// Withdraw LP Token from Onsen Pool
+                    IMasterChef(masterChef).withdrawAndHarvest(onsenPoolId, amount, address(this));
+                }
+            }
+        }
+    }
+
+    function _deactivate(bool panic) internal override {
+        /// Get amounts by LP to unstake and then deallocate (unstake) each token.
+        _deallocateAll();
```
PARABRAHMAN0, line 123: we can actually remove it in this case and just set all of it under panic

contracts/allocators/OnsenAllocatorV2.sol
```diff
@@ -0,0 +1,268 @@
...
+    mapping(address => PoolData) internal _lpToOnsenId;
+
+    /// ========== CONSTRUCTOR ==========
+
+    constructor(
+        address _chef,
+        address _sushi,
+        address _xSushi,
+        address _treasury,
+        AllocatorInitData memory data
```
PARABRAHMAN0, line 41: What if we want to add more tokens? Add a setter for more LP tokens, should be easy.

PARABRAHMAN0 (changes requested):

contracts/allocators/OnsenAllocatorV2.sol
calusaca: Function removed.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: Added the following 2 functions:

function setLPTokenOnsenPoolId(address lpToken, uint256 onsenPoolId) external onlyGuardian 
function getLPTokenOnsenPoolId(address lpToken) external view returns (bool, uint256)

So we can add the Onsen Pool Id for the LP Tokens ahead of time and also check if the LP token was already added, worst-case scenario would be that the token is not added and it will go search for it based on the Onsen Pool Length.

@Paul4912 Please let me know if this is fine or if we need to be able to add all Onsen Pool Ids at once?

contracts/allocators/OnsenAllocatorV2.sol
calusaca: image

I wasn't sure about this one since Visual Studio is suggesting me to make it "view", because it is finding that the only variable that is changing was declared inside the function, but it is actually changing a state variable so yes, I just change it.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: "Update" function updated so if there is no LP token balance, then harvest the sushi token rewards so those can be staked.

@Paul4912 @PARABRAHMAN0 , please check the current solution and let me know if you think this is fine.

Paul4912: yea that works too

contracts/allocators/OnsenAllocatorV2.sol
Paul4912: Yea that's how id imagine we'd do it. 

Idk how many onsen farms there are to need a batch add.

calusaca: @PARABRAHMAN0 @ind-igo Do you have any idea about how many Onsen Pools are used? Or do you know where I can check this? 
We would like to know if a batch function to add Onsen pools is needed or not?

contracts/allocators/OnsenAllocatorV2.sol
calusaca: Yes, you are right, the function was updated to call the withdrawAndHarvest in this same loop, avoiding looping it again.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: Function name updated.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: _deactivate function updated.

contracts/allocators/OnsenAllocatorV2.sol
calusaca: @PARABRAHMAN0  Sorry but not sure about what this means, or maybe is already added based on a previous comment with the following 2 functions?
function setLPTokenOnsenPoolId(address lpToken, uint256 onsenPoolId) external onlyGuardian
function getLPTokenOnsenPoolId(address lpToken) external view returns (bool, uint256)

calusaca:
> 

Sure, I thought I needed to add it as new PR, but sure I will push the testing to the branch and re-request review

philipjonsen (approved):

PR #281, opened 2022-03-31
added contract and test cases that let OHM treasury migrate it lps fr…

c-n-o-t-e:
…om sushiswap to uniswapv2

### DIFF
--- a/contracts/migration/SushiMigrator.sol
+++ b/contracts/migration/SushiMigrator.sol
@@ -30,12 +30,14 @@ contract SushiMigrator is OlympusAccessControlledV2 {
 
     constructor(address _olympusAuthority) OlympusAccessControlledV2(IOlympusAuthority(_olympusAuthority)) {}
 
-    // / @notice Takes lp from treasury, remove liquidity from sushi, adds liquidity to uniswap v2
-    // / @param sushiRouter_ sushi router addres
-    // / @param uniRouter_ gelato router address
-    // / @param sushiLpAddress_ lp address
-    // / @param amount_ lp amount to get from treasury and amount of liquidity to remove
-    // / @param percentage_ minimum percentage when using the addLiquidityGuni function....i.e 95% will be 950
+    /// @notice Takes lp from treasury, remove liquidity from sushi, adds liquidity to uniswap v2
+    /// @param sushiRouter_ sushi router addres
+    /// @param uniRouter_ uni router address
+    /// @param sushiLpAddress_ sushi lp address
+    /// @param uniswapLpAddress_ uni lp address
+    /// @param amount_ lp amount to get from treasury and amount of liquidity to remove
+    /// @param removeLiquiditySlippage_ slippage used when removing liquidity....i.e 95% will be 950
+    /// @param addLiquiditySlippage_ slippage used when adding liquidity....i.e 95% will be 950
     function executeTx(
         address sushiRouter_,
         address uniRouter_,
@@ -80,10 +82,11 @@ contract SushiMigrator is OlympusAccessControlledV2 {
         txCount++;
     }
 
-    // / @notice Removes liquidity from sushiswap
-    // / @param pairAddr_ lp address
-    // / @param router_ sushiswaprouter address
-    // / @param amount_ amount of lp to remove
+    /// @notice Removes liquidity from sushiswap
+    /// @param pairAddr_ lp address
+    /// @param router_ sushiswaprouter address
+    /// @param amount_ amount of lp to remove
+    /// @param slippage_ slippage to use
     function removeLiquidity(
         address pairAddr_,
         address router_,
@@ -111,12 +114,13 @@ contract SushiMigrator is OlympusAccessControlledV2 {
         );
     }
 
-    // / @notice Adds liquidity to uniswap pool
-    // / @param router_ uniswap router address
-    // / @param token0_ token address
-    // / @param token1_ token address
-    // / @param contractToken0Bal_ max amount of token 0 to be added as liquidity
-    // / @param contractToken1Bal_ max amount of token 1 to be added as liquidity
+    /// @notice Adds liquidity to uniswap pool
+    /// @param router_ uniswap router address
+    /// @param token0_ token address
+    /// @param token1_ token address
+    /// @param contractToken0Bal_ max amount of token 0 to be added as liquidity
+    /// @param contractToken1Bal_ max amount of token 1 to be added as liquidity
+    /// @param slippage_ slippage to use
     function addLiquidity(
         address router_,
         address token0_,
--- a/contracts/migration/SushiMigrator.sol
+++ b/contracts/migration/SushiMigrator.sol
@@ -1,0 +1,185 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity 0.8.10;
+
+import "../libraries/SafeERC20.sol";
+import "../interfaces/ITreasury.sol";
+import "../interfaces/IUniswapV2Pair.sol";
+import "../interfaces/IUniswapV2Router.sol";
+import "../types/OlympusAccessControlledV2.sol";
+
+/// @notice This contract handles migrating lps from uniswap v2 and sushiswap to uniswap v3 via gelato
+///         to make use of the contract ensure a pool has been created on uniswap v3 and gelato
+contract SushiMigrator is OlympusAccessControlledV2 {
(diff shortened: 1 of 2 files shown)


contracts/migration/SushiMigrator.sol
```diff
@@ -0,0 +1,189 @@
...
+        address uniswapLpAddress_,
+        uint256 amount_,
+        uint256 removeLiquiditySlippage_,
+        uint256 addLiquiditySlippage_
+    ) external onlyGovernor {
+        ITreasury(treasury).manage(sushiLpAddress_, amount_);
+
+        uint256 amount = IUniswapV2Pair(sushiLpAddress_).balanceOf(address(this));
+
+        (uint256 amountOHM, uint256 amountToken) = removeLiquidity(
```
PARABRAHMAN0, line 54: Is token0 always OHM? I know that this is not relevant to how the transaction develops further but it induces confusion too.

contracts/migration/SushiMigrator.sol
```diff
@@ -0,0 +1,189 @@
...
+import "../interfaces/IUniswapV2Pair.sol";
+import "../interfaces/IUniswapV2Router.sol";
+import "../types/OlympusAccessControlledV2.sol";
+
+/// @notice This contract handles migrating lps from sushiswap to uniswap v2
+///         to make use of this contract ensure contract address is RESERVEMANAGER IN OHM Treasury
+contract SushiMigrator is OlympusAccessControlledV2 {
+    using SafeERC20 for IERC20;
+
+    struct TxDetails {
```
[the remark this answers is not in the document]
PARABRAHMAN0: Also since this has to be read later the naming should be important here. A/B intermixed with the other is not explicit.  I would also change order. This would be good I think:
```
struct Amounts {
uint128 sushiLpBeforeMigration;
uint128 leftoverSushiLpAfterMigration;
uint128 uniPoolToken0AddedToPool;
uint128 uniPoolToken1AddedToPool;
uint128 uniPoolLpReceived;
uint128 uniPoolToken0ReturnedToTreasury;
uint256 uniPoolToken1ReturnedToTreasury;
}
```

Check below mapping.

contracts/migration/SushiMigrator.sol
```diff
@@ -0,0 +1,189 @@
...
+        uint256 contractV2lpBalanceBeforeRemovingLiquidity;
+        uint256 contractV2lpBalanceAfterRemovingLiquidity;
+        uint256 expectedToken0ToBeAddedOnUniswapV2;
+        uint256 expectedToken1ToBeAddedOnUniswapV2;
+        uint256 lp;
+    }
+
+    uint256 public txCount;
+
+    mapping(uint256 => TxDetails) public transactions;
```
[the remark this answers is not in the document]
PARABRAHMAN0: Also, I would advise to change the naming definitely to:

```
mapping(uint256 => Amounts) public amountsByMigrationId;
// or
Amounts[] public amountsByMigrationId;
```

contracts/migration/SushiMigrator.sol
```diff
@@ -0,0 +1,189 @@
...
+        (uint256 amountA, uint256 amountB, uint256 liquidity) = addLiquidity(
+            uniRouter_,
+            token0,
+            token1,
+            amountOHM,
+            amountToken,
+            addLiquiditySlippage_
+        );
+
+        transactions[txCount] = TxDetails({
```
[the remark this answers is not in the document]
PARABRAHMAN0: And yeah it would be amountsByMigrationId when changing naming above.

test/migration/SushiMigrator.js
```diff
@@ -0,0 +1,166 @@
...
+        await sushiMigrator
+            .connect(governor)
+            .executeTx(
+                sushiRouter,
+                uniRouter,
+                sushiOhmDaiLpAddress,
+                uniOhmDaiLpAddress,
+                amount,
+                990,
+                900
```
PARABRAHMAN0, line 64: It doesn't work with less than 10% slippage, so like 10% of underlying doesn't get added to LP?

test/migration/SushiMigrator.js
```diff
@@ -0,0 +1,166 @@
...
+            .connect(user1)
+            .transfer("0xa94A3Cd8326EA720f5dE5718C5519f409aFdb9c2", "5000000000000000000");
+
+        await user1.sendTransaction({
+            to: user.address,
+            value: ethers.utils.parseEther("2.0"), // Sends exactly 1.0 ether
+        });
+        await ohmContract.connect(user).approve(uniRouter, "1247350");
+
+        // The current price in the OHM/ETH uniswap pool isn't correct, using swapExactTokensForTokens to set the right price before proceeding.
```
PARABRAHMAN0, line 120: This is something that anybody executing will know I suppose.

PARABRAHMAN0: If it will still apply then.

PARABRAHMAN0 (changes requested):
Mostly naming changes and asked some questions. So if it's just a question or something not to be change you can just close it otherwise pls do the naming. Logic is cool and fine.

contracts/migration/SushiMigrator.sol
```diff
@@ -0,0 +1,189 @@
...
+            address,
+            address,
+            uint256,
+            uint256
+        )
+    {
+        address token0 = IUniswapV2Pair(lp_).token0();
+        address token1 = IUniswapV2Pair(lp_).token1();
+
+        uint256 token0Bal = IERC20(token0).balanceOf(addr_);
```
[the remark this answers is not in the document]
c-n-o-t-e: I left this the way it was because it was called in two places and both don't take lp address as a second parameter if that's okay.

contracts/migration/SushiMigrator.sol
```diff
@@ -0,0 +1,189 @@
...
+    /// @param router_ sushiswaprouter address
+    /// @param amount_ amount of lp to remove
+    /// @param slippage_ slippage to use
+    function removeLiquidity(
+        address pairAddr_,
+        address router_,
+        uint256 amount_,
+        uint256 slippage_
+    ) internal returns (uint256 amountOHM, uint256 amountToken) {
+        (address token0, address token1, uint256 pairBalanaceInTokenA, uint256 pairBalanaceInTokenB) = getTokenInfo(
```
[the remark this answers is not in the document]
c-n-o-t-e: Changed.

contracts/migration/SushiMigrator.sol
c-n-o-t-e: Yes in this case, given this is focus on OHM/DAI and OHM/ETH pools.

test/migration/SushiMigrator.js
c-n-o-t-e: Yes, which is sent back to the treasury. The percentage may vary depending on the reserve tokens amount.

PARABRAHMAN0 (changes requested):
Ignore this request.

PARABRAHMAN0 (approved):
Looks good to go.

contracts/migration/SushiMigrator.sol
```diff
@@ -0,0 +1,189 @@
...
+        address uniRouter_,
+        address sushiLpAddress_,
+        address uniswapLpAddress_,
+        uint256 amount_,
+        uint256 removeLiquiditySlippage_,
+        uint256 addLiquiditySlippage_
+    ) external onlyGovernor {
+        ITreasury(treasury).manage(sushiLpAddress_, amount_);
+
+        uint256 amount = IUniswapV2Pair(sushiLpAddress_).balanceOf(address(this));
```
JExtor, line 52: Should be okay to just use `amount_`

contracts/migration/SushiMigrator.sol
```diff
@@ -0,0 +1,189 @@
...
+
+        (uint256 amountOHM, uint256 amountToken) = removeLiquidity(
+            sushiLpAddress_,
+            sushiRouter_,
+            amount,
+            removeLiquiditySlippage_
+        );
+        uint256 amountAfterTx = IUniswapV2Pair(sushiLpAddress_).balanceOf(address(this));
```
JExtor, line 60: Shouldn't need this, all of `amount_` should be removed, should not be any left over liquidity

JExtor (commented):
Looks good!

ind-igo (approved):

PR #289, opened 2022-04-05
modify YieldSplitter to allow aggregator to redeem on behalf of recip…

Paul4912:
…ients

### DIFF
--- a/contracts/peripheral/YieldDirector.sol
+++ b/contracts/peripheral/YieldDirector.sol
@@ -304,7 +304,7 @@ contract YieldDirector is IYieldDirector, YieldSplitter {
         @notice Get redeemable gOHM balance of a recipient address
         @param recipient_ Address of user receiving donated yield
      */
-    function totalRedeemableBalance(address recipient_) public view override returns (uint256) {
+    function totalRedeemableBalance(address recipient_) external view override returns (uint256) {
         uint256[] memory receiptIds = recipientIds[recipient_];
 
         uint256 agnosticRedeemable = 0;
--- a/contracts/peripheral/YieldDirector.sol
+++ b/contracts/peripheral/YieldDirector.sol
@@ -7,7 +7,6 @@ import {IStaking} from "../interfaces/IStaking.sol";
 import {IYieldDirector} from "../interfaces/IYieldDirector.sol";
 import {SafeERC20} from "../libraries/SafeERC20.sol";
 import {YieldSplitter} from "../types/YieldSplitter.sol";
-import {OlympusAccessControlled, IOlympusAuthority} from "../types/OlympusAccessControlled.sol";
 
 /**
     @title  YieldDirector (codename Tyche) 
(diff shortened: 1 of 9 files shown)


contracts/peripheral/YieldDirector.sol
```diff
@@ -314,7 +304,7 @@ contract YieldDirector is IYieldDirector, YieldSplitter, OlympusAccessControlled
         @notice Get redeemable gOHM balance of a recipient address
         @param recipient_ Address of user receiving donated yield
      */
-    function totalRedeemableBalance(address recipient_) external view override returns (uint256) {
+    function totalRedeemableBalance(address recipient_) public view override returns (uint256) {
```
0xLienid, line 307: should be set to external for gas

contracts/peripheral/YieldDirector.sol
```diff
@@ -287,15 +286,6 @@ contract YieldDirector is IYieldDirector, YieldSplitter, OlympusAccessControlled
         return _getOutstandingYield(principalTotal, agnosticTotal);
     }
 
-    /**
-        @notice Getter function for a depositor's list of IDs. This is needed for the frontend
-                as public state variables that map to arrays only return one element at a time
-                rather than the full array
-    */
-    function getDepositorIds(address donor_) external view override returns (uint256[] memory) {
```
0xLienid, line 295: this should stay, returns the full array whereas a public variable doesn't

contracts/peripheral/YieldDirector.sol
```diff
@@ -427,17 +443,18 @@ contract YieldDirector is IYieldDirector, YieldSplitter, OlympusAccessControlled
     /**
         @notice Redeem available gOHM yield from a specific deposit
         @param depositId_ Deposit ID to withdraw gOHM yield from
+        @param recipient_ address of recipient
     */
-    function _redeem(uint256 depositId_) internal returns (uint256 amountRedeemed) {
+    function _redeem(uint256 depositId_, address recipient_) internal returns (uint256 amountRedeemed) {
         if (redeemDisabled) revert YieldDirector_RedeemsDisabled();
-        if (recipientLookup[depositId_] != msg.sender) revert YieldDirector_NotYourYield();
+        if (recipientLookup[depositId_] != recipient_) revert YieldDirector_NotYourYield();
```
0xLienid, line 450: I think this tracks as still being protected from someone else redeeming your yield, was it thoroughly tested though?

contracts/peripheral/YieldDirector.sol
0xLienid: nevermind, I see it's added to YieldSplitter

test/tyche/YieldDirector.test.js
```diff
@@ -505,6 +515,21 @@ describe("YieldDirector", async () => {
         await expect(bobBalance).is.equal(donatedAmount.add("2"));
     });
 
+    it("should redeem all tokens on bahalf of others", async () => {
```
0xLienid, line 518: should we also add a test to make sure it reverts if .givePermissionToRedeem isn't called?

0xLienid (approved):

contracts/peripheral/YieldDirector.sol
Paul4912: So i changed those internal functions to use parameter recipient because we now have a function that allows external parties to redeem on behalf of users. we want to use passed in recipient rather than msg.sender. In most cases redeemer is msg.sender, but it will be passed in.

test/tyche/YieldDirector.test.js
Paul4912: done

PR #291, opened 2022-04-09
feat: OlympusChainlinkBasedSwapper

PARABRAHMAN0:
This contract is a permissionless, accessless, swapper contract using Chainlink datafeeds to allow any manners of swaps for us or anyone else to execute. It allows swaps to be executed to UNI V3 via their SwapRouter v1 and to any other forks of UNI V2 + UNI V2 over the other router. The plan is to use this for swapping CRV to CVX after using the Curve pool with low slippage set (1:1 peg we don't care about price) to swap from cvxCRV to CRV. After this we use Uniswap. On Uniswap V3 I received good results with 97% slippage in tests [not shown in PR].

### DIFF
--- a/contracts/accessless/OlympusChainlinkBasedSwapper.sol
+++ b/contracts/accessless/OlympusChainlinkBasedSwapper.sol
@@ -1,0 +1,215 @@
+pragma solidity ^0.8.7;
+
+import "@chainlink/contracts/src/v0.8/interfaces/FeedRegistryInterface.sol";
+import "@chainlink/contracts/src/v0.8/Denominations.sol";
+import "@uniswap/v3-periphery/contracts/interfaces/ISwapRouter.sol";
+import "@uniswap/v2-periphery/contracts/interfaces/IUniswapV2Router02.sol";
+import "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+
+struct V3Params {
+    uint24[] fees;
+    address[] path;
+    address denomination;
+    address recipient;
+    uint256 deadline;
+    uint256 amount;
+    uint256 slippage;
+}
+
+struct V2Params {
+    address router;
+    address[] path;
+    address denomination;
+    address recipient;
+    uint256 deadline;
+    uint256 amount;
+    uint256 slippage;
+}
+
+error OlympusChainlinkBasedSwapper_WrongV3PathLength();
+error OlympusChainlinkBasedSwapper_ApproveFailed();
+error OlympusChainlinkBasedSwapper_TransferFromFailed();
+error OlympusChainlinkBasedSwapper_TransferFailed();
+
+contract OlympusChainlinkBasedSwapper {
+    address public constant weth = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+    address public constant wbtc = 0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599;
+
+    FeedRegistryInterface public registry;
+    ISwapRouter public v3SwapRouter;
+
+    constructor(address _registry, address _v3SwapRouter) {
+        registry = FeedRegistryInterface(_registry);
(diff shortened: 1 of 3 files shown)


PARABRAHMAN0:
Added missing dependencies but tests of other allocators will fail because extender id is incremented and conflicts with hardcoded.

PARABRAHMAN0:
- state variables can be immutable, noticed deliberating whether to commit or leave as is since deployed

contracts/accessless/OlympusChainlinkBasedSwapper.sol
```diff
@@ -0,0 +1,215 @@
...
+    constructor(address _registry, address _v3SwapRouter) {
+        registry = FeedRegistryInterface(_registry);
+        v3SwapRouter = ISwapRouter(_v3SwapRouter);
+    }
+
+    function v3ExactInput(V3Params calldata params) external returns (uint256 amountOut) {
+        amountOut = _exact(params, true);
+    }
+
+    function v3ExactOutput(V3Params calldata params) external returns (uint256 amountIn) {
```
0xLienid, line 52: is there a reason to not have v3ExactInput and v3ExactOutput be one function with an additional boolean parameter to signify input vs output? and then just call `amount = _exact(params, new_boolean_param)`?

contracts/accessless/OlympusChainlinkBasedSwapper.sol
```diff
@@ -0,0 +1,215 @@
...
+        address denomination = (params.denomination == address(0)) ? Denominations.USD : params.denomination;
+
+        // interactions + checks
+        // didn't abstract to keep readable
+
+        bool check = IERC20(inputToken).transferFrom(msg.sender, address(this), params.amount);
+
+        if (!check) revert OlympusChainlinkBasedSwapper_TransferFromFailed();
+
+        check = IERC20(inputToken).approve(params.router, params.amount);
```
0xLienid, line 70: does approve work on objects like this rather than addresses? very good to know

contracts/accessless/OlympusChainlinkBasedSwapper.sol
```diff
@@ -0,0 +1,215 @@
...
+            );
+        }
+
+        // cleanup
+        _clean(inputToken);
+    }
+
+    // @notice
+    function _clean(address inputToken) internal {
+        bool check = IERC20(inputToken).transfer(msg.sender, IERC20(inputToken).balanceOf(address(this)));
```
0xLienid, line 150: are there edge cases where transferring the full balance is not the desired behavior?

contracts/accessless/OlympusChainlinkBasedSwapper.sol
```diff
@@ -0,0 +1,215 @@
...
+        uint256 inPrice = uint256(_latestPrice(inputToken, denom));
+        uint256 outPrice = uint256(_latestPrice(outputToken, denom));
+
+        // slippage must be normalized to 1e18
+        amountMin = (amount * inPrice * slippage) / (1e18 * outPrice);
+        amountMin = (amountMin * IERC20Metadata(outputToken).decimals()) / IERC20Metadata(inputToken).decimals();
+    }
+
+    function _latestPrice(address base, address quote) internal view returns (int256 price) {
+        if (base == weth) base = Denominations.ETH;
```
0xLienid, line 170: might be cleaner to do `base = base == weth ? Denominations.ETH : Denominations.BTC;`

contracts/accessless/OlympusChainlinkBasedSwapper.sol
PARABRAHMAN0: yes to make it explicit

contracts/accessless/OlympusChainlinkBasedSwapper.sol
PARABRAHMAN0: what? IERC20(inputToken) == IERC20, params.router == address, amount == uint256

contracts/accessless/OlympusChainlinkBasedSwapper.sol
PARABRAHMAN0: Tokens should never be stuck in the contract, so always transfer out. It should never happen that your tokens get locked inside it.

PR #295, opened 2022-04-15
Incur Debt: Convert internal accounting to use gOHM.

Paul4912:
Tests needs to be updated.

### DIFF
--- a/contracts/peripheral/IncurDebt.sol
+++ b/contracts/peripheral/IncurDebt.sol
@@ -33,10 +33,7 @@ contract IncurDebt is OlympusAccessControlledV2, IIncurDebt {
     event BorrowerRevoked(address indexed _borrower, bool _lpBorrower, bool _nonLpBorrower);
     event BorrowerDebtLimitSet(address indexed _borrower, uint256 _limit);
 
-    event BorrowerDeposit(
-        address indexed _borrower,
-        uint256 _amountToDeposit
-    );
+    event BorrowerDeposit(address indexed _borrower, uint256 _amountToDeposit);
     event Borrowed(
         address indexed _borrower,
         uint256 _amountToBorrow,
@@ -81,17 +78,14 @@ contract IncurDebt is OlympusAccessControlledV2, IIncurDebt {
         uint256 _collateralLeftForWithdraw
     );
 
-    event Withdrawal(
-        address indexed _borrower,
-        uint256 _amountToWithdraw,
-        uint256 _currentCollateral
-    );
+    event Withdrawal(address indexed _borrower, uint256 _amountToWithdraw, uint256 _currentCollateral);
 
     uint256 public globalDebtLimit;
     uint256 public totalOutstandingGlobalDebt;
 
     address public immutable OHM;
     address public immutable gOHM;
+    address public immutable sOHM;
     address public immutable staking;
     address public immutable treasury;
 
@@ -111,16 +105,19 @@ contract IncurDebt is OlympusAccessControlledV2, IIncurDebt {
     constructor(
         address _OHM,
         address _gOHM,
+        address _sOHM,
         address _staking,
         address _treasury,
         address _olympusAuthority
     ) OlympusAccessControlledV2(IOlympusAuthority(_olympusAuthority)) {
         OHM = _OHM;
         gOHM = _gOHM;
+        sOHM = _sOHM;
         staking = _staking;
         treasury = _treasury;
         IERC20(OHM).safeApprove(treasury, type(uint256).max);
         IERC20(gOHM).safeApprove(staking, type(uint256).max);
+        IERC20(sOHM).safeApprove(staking, type(uint256).max);
     }
 
     modifier isBorrower(address _borrower) {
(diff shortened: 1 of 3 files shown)


c-n-o-t-e (commented):
I will share details via call

contracts/peripheral/IncurDebt.sol
```diff
@@ -443,60 +430,67 @@ contract IncurDebt is OlympusAccessControlledV2, IIncurDebt {
...
+
+        ITreasury(treasury).incurDebt(_ohmAmount, OHM);
     }
 
     /// @notice pays borrower debit and return left collateral and paid debt
     /// @param _borrower the account debit needs to be paid
-    /// @return currentCollateral left collateral after paying debit
+    /// @return collateralRemaining left collateral after paying debit
     /// @return paidDebt debt paid
     function _repay(address _borrower)
```
c-n-o-t-e, line 474: Kindly add something of this sort

```
function _repay(address _borrower, bool _seize)………
if(!_seize){
	uint256 sOHMToWrap = IgOHM(gOHM).balanceFrom(borrower.unwrappedGOHM) - debt;
	borrower.unwrappedGOHM = 0;
	IStaking(staking).wrap(address(this), sOHMToWrap); // wrap up remaining sOHM into gOHM
}
```

contracts/peripheral/IncurDebt.sol
```diff
@@ -242,17 +240,14 @@ contract IncurDebt is OlympusAccessControlledV2, IIncurDebt {
...
     function seize(address _borrower) external override onlyGovernor isBorrower(_borrower) {
         (uint256 seizedCollateral, uint256 paidDebt) = _repay(_borrower);
+        borrowers[_borrower].collateralInGOHM = 0;
 
-        clearBorrower(_borrower);
-
-        IERC20(sOHM).safeApprove(staking, seizedCollateral);
-        IStaking(staking).unstake(address(this), seizedCollateral, false, true);
-        IOHM(OHM).burn(seizedCollateral);
+        IgOHM(gOHM).burn(address(this), seizedCollateral);
```
c-n-o-t-e, line 245: Only staking contract can burn ohm
Unstake borrowers sohm to ohm and burn

contracts/peripheral/IncurDebt.sol
```diff
@@ -383,23 +367,27 @@ contract IncurDebt is OlympusAccessControlledV2, IIncurDebt {
...
-    /// @param _amount amount of sOHM to withdraw
-    /// @param _to address to send _amount
-    function withdraw(uint256 _amount, address _to) external override isBorrower(msg.sender) {
+    /// @param _amount amount of gOHM to withdraw
+    function withdraw(uint256 _amount) external override isBorrower(msg.sender) {
         if (_amount == 0) revert IncurDebt_InvaildNumber(_amount);
         Borrower storage borrower = borrowers[msg.sender];
 
-        updateCollateralInSOHM();
         if (_amount > getAvailableToBorrow()) revert IncurDebt_AmountAboveBorrowerBalance(_amount);
```
c-n-o-t-e, line 378: Amount is gohm, getAvailableToBorrow() returns borrowers balance in sohm.

contracts/peripheral/IncurDebt.sol
```diff
@@ -383,23 +367,27 @@ contract IncurDebt is OlympusAccessControlledV2, IIncurDebt {
...
 
-        borrower.collateralInSOHM -= uint128(_amount);
-        borrower.borrowerGonBalance -= IsOHM(sOHM).gonsForBalance(_amount);
-        IERC20(sOHM).transfer(msg.sender, _amount);
+        if (_amount > borrower.collateralInGOHM - borrower.unwrappedGOHM) { // Does uint256 > uint128 cause problem?
+            uint256 amountGOHMToWrap = borrower.unwrappedGOHM + _amount - borrower.collateralInGOHM;
+            borrower.unwrappedGOHM -= uint128(amountGOHMToWrap);
```
c-n-o-t-e, line 382: This will always fail say;
`borrower.collateralInGOHM` = 10 gOHM
`borrower.unwrappedGOHM` = 5 gOHM
`amount` = 11 gOHM
The calculation from line `381` will equal to 6
Resulting to 5 -6

contracts/peripheral/IncurDebt.sol
Paul4912: So for seize im unstaking the gOHM to OHM to burn anyway. If we do this and repay doesn't wrap up excess sOHM we can run into situation where we have some collateral in sOHM some in gOHM which makes burning bit more difficult logic. So rather just wrap up all excess sOHM and then in burn function unstake and burn it. Also cleaner for the repay function.

contracts/peripheral/IncurDebt.sol
Paul4912: Not sure how to rename variables to make this clearer but
collateralInGOHM is meant to be their total collateral value in gOHM terms
unwrappedGOHM is meant to represent how much of their collateralInGOHM is unwrapped in sOHM. 

so unwrappedGOHM <= collateralInGOHM

so max amount is = collateralInGOHM which is >= unwrappedGOHM

contracts/peripheral/IncurDebt.sol
Paul4912: done

contracts/peripheral/IncurDebt.sol
Paul4912: ahh yes

c-n-o-t-e (approved):

PR #296, opened 2022-04-15
Update StakingDistributor.sol

olyzeus:
Updated distributor adds the ability to mint and sync into Uniswap V2-style liquidity pools, removing the opportunity-cost dilemma of providing liquidity for OHM, as well as patches a small bug in the staking contract that pulls forward an amount of the next epoch rewards. Note that this implementation bases staking reward distributions on staked supply, whereas previous implementations used total supply.


contracts/StakingDistributor.sol
```diff
@@ -1,211 +1,200 @@
 // SPDX-License-Identifier: AGPL-3.0
-pragma solidity ^0.7.5;
-
-import "./libraries/SafeERC20.sol";
-import "./libraries/SafeMath.sol";
+pragma solidity ^0.8.10;
+pragma abicoder v2;
```
Oighty, line 3: Don't need abicoder v2 in solidity 0.8. It's the default (see https://docs.soliditylang.org/en/v0.8.13/080-breaking-changes.html#silent-changes-of-the-semantics)

contracts/StakingDistributor.sol
```diff
@@ -1,211 +1,200 @@
...
-            require(_rate <= info[_index].rate.mul(25).div(1000), "Limiter: cannot adjust by >2.5%");
-        }
-
-        if (!_add) {
-            require(_rate <= info[_index].rate, "Cannot decrease rate by more than it already is");
-        }
+    ) external {
+        if (msg.sender != authority.governor() && msg.sender != authority.guardian()) revert Not_Permissioned();
+        if (msg.sender == authority.guardian() && _rate > rewardRate * 25 / 1000) revert Adjustment_Limit();
+        if (!_add && _rate > rewardRate) revert Adjustment_Overflow();
```
Oighty, line 196: Error would be an underflow instead of an overflow (kind of a nitpick).

contracts/StakingDistributor.sol
```diff
@@ -1,211 +1,200 @@
...
+    /// The OHM Staking Contract
     address private immutable staking;
 
-    mapping(uint256 => Adjust) public adjustments;
-    uint256 public override bounty;
-
-    uint256 private immutable rateDenominator = 1_000_000;
+    /// The % to increase balances per epoch
+    uint256 public rewardRate;
+    /// Liquidity pools to receive rewards
```
Oighty, line 45: Not necessarily relevant to getting this out for M&S, but there may be other cases in the future where a contract is holding naked OHM and could/should? receive rebases. You could generalize this to "receivers", but the interfaces may not conform to the "sync".

contracts/StakingDistributor.sol
```diff
@@ -1,211 +1,200 @@
...
 
-contract Distributor is IDistributor, OlympusAccessControlled {
-    /* ========== DEPENDENCIES ========== */
+/// @notice Updated distributor adds the ability to mint and sync
+///         into Uniswap V2-style liquidity pools, removing the
+///         opportunity-cost dilemma of providing liquidity for
+///         OHM, as well as patches a small bug in the staking contract
+///         that pulls forward an amount of the next epoch rewards. Note that
+///         this implementation bases staking reward distributions on
+///         staked supply, whereas previous implementations used total supply.
```
Oighty, line 18: This can be clarified to say "bases staking reward distributions on staked and LP supply".

Oighty (commented):
I don't see any major issues. I made a few suggestions on comments and a couple minor changes.

contracts/StakingDistributor.sol
olyzeus: implemented

contracts/StakingDistributor.sol
olyzeus: we might want to implement this this functionality on the heart or tangential contract. i dont think we would need it before then

contracts/StakingDistributor.sol
olyzeus: save it for the semantics dome, E.B. White

ind-igo (approved):

PR #297, opened 2022-04-18
Rewrote tests based on mint and sync and on forked mainnet

0xLienid:
I also added an `_initialRate` parameter to the constructor as it seemed that with the removal of being able to set rates by recipient through `addRecipient` and the existence of the check at Line 196, rewardRate was perpetually stuck at 0. If I simply missed the right way to set rewardRate, lmk and I can remove that parameter.


test/staking/StakingDistributor.test.ts
```diff
@@ -9,341 +9,395 @@ import {
...
 
-            it("can only be called by the staking contract", async () => {
+            it("mints to pools", async () => {
+                await advanceEpoch();
+                const pools = ["0x055475920a8c93cffb64d039a8205f7acc7722d3"];
+                await distributor.connect(governor).setPools(pools);
+
+                let [reserve0, reserve1, timestamp] = await ohmDai.getReserves();
+                const priceBefore = reserve1 / (reserve0 * 1000000000);
+                const balanceBefore = await ohm.balanceOf(pools[0]);
```
olyzeus, line 138: can you add a expectedBalanceAfter = balanceBefore * rewardRate ?

everything else looks good, just a good sanity check here rather than vaguer >

ind-igo (approved):

PR #308, opened 2022-04-26
remove slippage and use minAmount

Paul4912:
Think this is simpler for others to use raw min amount instead of having to calculate slippage and min amount. No need for both.


contracts/peripheral/Strategies/UniSwap.sol
```diff
@@ -60,9 +60,8 @@ contract UniSwapStrategy is IStrategy {
             uint256 amountADesired,
             uint256 amountBDesired,
             uint256 amountAMin,
-            uint256 amountBMin,
-            uint256 slippage
```
c-n-o-t-e, line 64: Everything would work well without slippage but not every time which is tied to reserves of tokens A and B. So let's keep that so that if such a case happens everything still works fine.

PR #318, opened 2022-06-06
<test/Allocators>: add test case for aave allocator

david30907d:
# WIP

1. Hi there, I'm wondering if writing unittest for allocators is a good use of time?
I'm solidity noob and want to try some easy ticket to ramp up~
2. How can I enable CI on my personal repo? I've checked [.github](https://github.com/OlympusDAO/olympus-contracts/blob/main/.github/workflows/CI.yml) but still have no idea why CI workflow was not triggered when PR created in my forked repo

### DIFF
--- a/test/allocators/AaveAllocator.test.ts
+++ b/test/allocators/AaveAllocator.test.ts
@@ -1,0 +1,81 @@
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/dist/src/signer-with-address";
+import chai, { expect } from "chai";
+import { ethers } from "hardhat";
+const { BigNumber } = ethers;
+import { FakeContract, smock } from "@defi-wonderland/smock";
+import {
+    IERC20,
+    IERC20Metadata,
+    ITreasury,
+    IStabilityPool,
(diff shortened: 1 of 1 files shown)


test/allocators/AaveAllocator.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+            describe("setter tests", () => {
+                it("invalid setReferralCode", async () => {
+                    try {
+                        await expect(aaveAllocator.connect(owner).setReferralCode(-1))
+                        .to.be.revertedWith("value out-of-bounds (argument=\"code\", value=-1, code=INVALID_ARGUMENT, version=abi/5.5.0)");
+                    } catch (error) {
+                        console.log(error)
+                    }
+                });
```
david30907d, line 79: forgive my ignorance (I'm solidity noob), how can I `expect` an exception to be thrown in unittest?

In python, I can do this to assert an exception get raised in unittest:
```python
def test_passes():
    with pytest.raises(Exception) as e_info:
        x = 1 / 0
```

Already googled around but didn't found useful hardhat doc

PR #334, opened 2022-10-10
OHM Bond Launch Wrapper

0xLienid:


ind-igo (approved):

contracts/peripheral/OhmBondManager.sol
```diff
@@ -0,0 +1,167 @@
...
+
+    // ========= MARKET CREATION ========= //
+    function createBondProtocolMarket(uint256 capacity_, uint256 bondTerm_) external onlyPolicy returns (uint256) {
+        _topUpOhm(capacity_);
+
+        /// Encodes the information needed for creating a bond market on Bond Protocol
+        bytes memory createMarketParams = abi.encode(
+            ohm, // payoutToken
+            ohm, // quoteToken
+            bondProtocolParameters.callbackAddress, // callbackAddress
```
Oighty, line 71: I would remove the callback option from this version. It doesn't make sense to use one in this context since the contract itself will be providing the OHM.

test/bonds/OhmBondManager.test.ts
```diff
@@ -0,0 +1,332 @@
...
+
+        guardian = await impersonate(await authority.guardian());
+        policy = await impersonate(await authority.policy());
+
+        ohm = await getCoin(coins.ohm);
+
+        ohmBondManager = await new OhmBondManager__factory(owner).deploy(
+            ohm.address,
+            olympus.treasury,
+            "0x007FEA7A23da99F3Ce7eA34F976f32BF79A09C43", // Fixed Expiry Auctioneer
```
Oighty, line 68: Need to update auctioneer address in test.

test/bonds/OhmBondManager.test.ts
```diff
@@ -0,0 +1,332 @@
...
+            "0x007FE7c498A2Cf30971ad8f2cbC36bd14Ac51156", // Fixed Expiry Teller
+            "0x0b7fFc1f4AD541A4Ed16b40D8c37f0929158D101", // Gnosis Easy Auction
+            authority.address
+        );
+
+        feAuctioneer = await ethers.getContractAt(
+            feAuctioneerAbi,
+            "0x007FEA7A23da99F3Ce7eA34F976f32BF79A09C43"
```
Oighty, line 76: Same as above.

contracts/peripheral/OhmBondManager.sol
```diff
@@ -0,0 +1,167 @@
...
+    ) OlympusAccessControlled(IOlympusAuthority(authority_)) {
+        ohm = IERC20(ohm_);
+        treasury = ITreasury(treasury_);
+        fixedExpiryAuctioneer = IBondSDA(feAuctioneer_);
+        fixedExpiryTeller = IBondTeller(feTeller_);
+        gnosisEasyAuction = IEasyAuction(gnosisAuction_);
+    }
+
+    // ========= MARKET CREATION ========= //
+    function createBondProtocolMarket(uint256 capacity_, uint256 bondTerm_) external onlyPolicy returns (uint256) {
```
Oighty, line 64: I'm a little torn between providing a duration and a timestamp for the bondTerm. It would be easier to line up several bonds to use the same token if you used a timestamp since it's a little uncertain when MS transactions will go through. However, this is more likely to result in the outcome expected for issuing a single bond.

Oighty (commented):
Overall, looks goods.

contracts/peripheral/OhmBondManager.sol
0xLienid: hmm yeah that's a good point I hadn't thought about. I was just coming at it from the perspective we want to minimize potential truncation of bond term if the multisig takes a while to execute.

PR #338, opened 2023-01-20
Aura Allocators V2

0xLienid:
Allocators to deposit BPT into Aura, and to manage vlAura.

AuraAllocatorV2 manages vlAura and is ready for review as it is higher priority than the balancer one


ind-igo (approved):

contracts/allocators/AuraAllocatorV2.sol
```diff
@@ -0,0 +1,236 @@
...
+
+    function delegate(address delegatee_) external onlyGuardian {
+        locker.delegate(delegatee_);
+    }
+
+    function setShouldLock(bool shouldLock_) external onlyGuardian {
+        shouldLock = shouldLock_;
+    }
+
+    function setLocker(address locker_) external onlyGuardian {
```
Oighty, line 196: Does this function assume that the existing locker doesn't hold anymore funds? If so, you may want to check that before setting new value.

contracts/allocators/AuraAllocatorV2.sol
```diff
@@ -0,0 +1,236 @@
...
+        aura.approve(address(locker), 0);
+        aura.approve(address(locker_), type(uint256).max);
+
+        locker = IAuraLocker(locker_);
+    }
+
+    function setAbStaking(address abStaking_) external onlyGuardian {
```
Oighty, line 210: Same as above.

PR #339, opened 2023-04-10
Add new allocator to migrate old LUSD allocator to and rescue funds

0xLienid:
Test CI check is failing because of some curve convex allocator test, nothing to do with this allocator

### DIFF
--- a/contracts/allocators/LUSDAllocatorRescue.sol
+++ b/contracts/allocators/LUSDAllocatorRescue.sol
@@ -1,0 +1,78 @@
+pragma solidity ^0.8.10;
+
+import "../interfaces/IAllocator.sol";
+import "./interfaces/IWETH.sol";
+import "../types/BaseAllocator.sol";
+
+error LUSDAllocator_InputTooLarge();
+error LUSDAllocator_TreasuryAddressZero();
+
+/**
+ *  Contract deploys LUSD from treasury into the liquity stabilty pool. Each update, rewards are harvested.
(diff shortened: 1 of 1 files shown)


contracts/allocators/LUSDAllocatorRescue.sol
```diff
@@ -0,0 +1,78 @@
...
+     * @notice tokens in AllocatorInitData should be [LUSD Address]
+     * LUSD Address (0x5f98805A4E8be255a32880FDeC7F6728C6568bA0)
+     */
+    constructor(AllocatorInitData memory data) BaseAllocator(data) {}
+
+    // Send ETH to an address
+    function sendETH(address to_) external {
+        _onlyGuardian();
+
+        payable(to_).transfer(address(this).balance);
```
Oighty: It's better to use a low-level call since if you send it to a smart contract wallet or a contract, it could get bricked in the future if the gas cost of op codes change. Transfer only forwards 2300 gas.

`payable(to_).call{value: address(this).balance}("");`

Oighty (commented):
Overall, looks fine. It's probably best to just do the rescue from the allocator vs. trying to migrate from this one to a new one.

contracts/allocators/LUSDAllocatorRescue.sol
0xLienid: cool, updated

0xLienid:
> Overall, looks fine. It's probably best to just do the rescue from the allocator vs. trying to migrate from this one to a new one.

yep agreed. i just wanted to have the ability to migrate again in case something unexpected goes wrong with this

